PR #2, opened 2016-11-27
Initial structure

aftersomemath:
This node is still a WIP. However the initial structure for development is here and proof of concept has been demonstrated. Thus initial structure should be considered done.


src/LowLevelMotionController.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+    ros::NodeHandle nh;
+
+    QuadVelocityController quadController(nh);
+
+    quadController.init();
+
+    while(ros::ok())
+    {
+        quadController.update();
```
amiller27: Don't we want this to be rate limited?  We don't want this hogging all our CPU when we only need it to run at under 100Hz

CMakeLists.txt
```diff
@@ -126,16 +129,16 @@ include_directories(
 # add_dependencies(iarc7_motion ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS})
 
 ## Declare a C++ executable
-# add_executable(iarc7_motion_node src/iarc7_motion_node.cpp)
+add_executable(LowLevelMotionController src/LowLevelMotionController.cpp src/FeedForwardPid.cpp src/QuadVelocityController.cpp)
```
amiller27: I've always seen CMake targets as lowercase_with_underscores in ROS projects, so I assumed that was the convention we were using, but we don't have to

src/QuadVelocityController.cpp
```diff
@@ -0,0 +1,115 @@
...
+yaw_pid_(0.0, 0.0, 0.0) ,
+tfBuffer_(),
+tfListener_(tfBuffer_)
+{
+
+}
+
+void QuadVelocityController::init()
+{
+    uav_control_ = nh_.advertise<iarc7_msgs::OrientationThrottleStamped>("uav_direction_command", 50);
```
amiller27: I'm assuming this isn't in the constructor because it can throw exceptions, and it isn't in a factory method because that would require dynamic allocation?

src/LowLevelMotionController.cpp
aftersomemath: It depends on QuadVelocityController.cpp line 77 where it looks up the latest transform in the buffer. That blocks and I think it might yield, I'll have to look into it more.

If that does not yield then the velocity code will need a bit of a refactoring to detect changes in the transform. At that point I would say that should be a new feature that doesn't need fixing in initial structure.

CMakeLists.txt
aftersomemath: Fixed

src/LowLevelMotionController.cpp
aftersomemath: I'll make a trello card if we need the feature.

amiller27: I think it's something we should look into if we aren't sure what the timing on this will be, because it would be good to know if there's latency or a low frequency update here because of waiting for transforms

aftersomemath: I have that by specifying a timeout for how long it will wait for a new transform before spitting out ROS_ERROR's.

Also it checks to see that the time between transforms isn't too long.

amiller27: Ok

src/QuadVelocityController.cpp
aftersomemath: No I needed init for something else I think, then I ended up not needing it as you pointed out. Advertise does not throw exceptions, so this will be moved to the constructor.

PR #6, opened 2016-12-06
Twist limiter

aftersomemath:
A class has been added to limit the max slew rate, min, and max requests that are made to the flight controller. QuadVelocityController was slightly refactored to remove responsibilities it shouldn't have had.

I'm using a rather ugly way to get the parameters into the new QuadTwistLimiter class however the comments explain how that will be cleaned up in the future with rosparam and switching QuadVelocityController to use twists.

### DIFF
--- a/include/iarc7_motion/QuadTwistRequestLimiter.hpp
+++ b/include/iarc7_motion/QuadTwistRequestLimiter.hpp
@@ -32,13 +32,13 @@ namespace Iarc7Motion
         QuadTwistRequestLimiter(const QuadTwistRequestLimiter& rhs) = delete;
         QuadTwistRequestLimiter& operator=(const QuadTwistRequestLimiter& rhs) = delete;
 
-        TwistStamped limitTwist(const TwistStamped& twist);
+        void limitTwist(TwistStamped& input_twist);
 
     private:
 
-        static double velocityLimit(double start, double end, double max, ros::Duration& delta, char const * axis);
-        static double minLimit(double request, double min, char const * axis);
-        static double maxLimit(double request, double max, char const * axis);
+        static void velocityLimit(double& request, const double old, const double max, const ros::Duration& delta, char const * axis);
+        static void minLimit(double& request, const double min, char const * axis);
+        static void maxLimit(double& request, const double max, char const * axis);
 
         Twist minTwist_;
         Twist maxTwist_;
--- a/include/iarc7_motion/QuadTwistRequestLimiter.hpp
+++ b/include/iarc7_motion/QuadTwistRequestLimiter.hpp
@@ -1,0 +1,49 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Twist Limiter
+//
+// Limits the rate of change of a twist and min/max values
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef QUAD_TWIST_LIMITER_HPP
+#define QUAD_TWIST_LIMITER_HPP
+
+#include <ros/ros.h>
+#include "geometry_msgs/Twist.h"
+#include "geometry_msgs/TwistStamped.h"
+
+using geometry_msgs::TwistStamped;
+using geometry_msgs::Twist;
+
+namespace Iarc7Motion
+{
+
+    class QuadTwistRequestLimiter
+    {
+    public:
+        QuadTwistRequestLimiter() = delete;
+
+        QuadTwistRequestLimiter(Twist minTwist, Twist maxTwist, Twist maxChange);
+
+        ~QuadTwistRequestLimiter() = default;
+
+        // Don't allow the copy constructor or assignment.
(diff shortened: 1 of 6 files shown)


include/iarc7_motion/QuadTwistRequestLimiter.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include "geometry_msgs/Twist.h"
+#include "geometry_msgs/TwistStamped.h"
+
+using geometry_msgs::TwistStamped;
+using geometry_msgs::Twist;
+
+namespace Iarc7Motion
+{
+
+    class QuadTwistRequestLimiter
```
amiller27: I thought we weren't indenting namespaces?

src/LowLevelMotionController.cpp
```diff
@@ -7,10 +7,41 @@
...
+#include "geometry_msgs/Twist.h"
+#include "geometry_msgs/TwistStamped.h"
 
 using namespace Iarc7Motion;
+using geometry_msgs::TwistStamped;
+using geometry_msgs::Twist;
+
+// This is a helper function that will limit a uav command using the twist limiter
+// It will eventually no longer be needed when the quad velocity controller is converted to use twists
+void limitUavCommand(QuadTwistRequestLimiter& limiter, iarc7_msgs::OrientationThrottleStamped& uav_command)
```
amiller27: Why won't this be needed?  Are we changing the output of low level motion to a twist?

src/LowLevelMotionController.cpp
```diff
@@ -24,9 +55,46 @@ int main(int argc, char **argv)
 
     AccelerationPlanner<QuadVelocityController> accelerationPlanner(nh, quadController);
 
+    ros::Publisher uav_control_ = nh.advertise<iarc7_msgs::OrientationThrottleStamped>("uav_direction_command", 50);
+
+    // Check for empty uav_control_ as per http://wiki.ros.org/roscpp/Overview/Publishers%20and%20Subscribers
+    // section 1
+    if(!uav_control_)
```
amiller27: I believe this if statement should just be `ROS_ASSERT_MSG(uav_control_, "Could not...");`

amiller27: Even if we still want the if statement, the body needs to change

src/QuadTwistRequestLimiter.cpp
```diff
@@ -0,0 +1,91 @@
...
+using namespace Iarc7Motion;
+QuadTwistRequestLimiter::QuadTwistRequestLimiter(Twist minTwist, Twist maxTwist, Twist maxChange) :
+minTwist_(minTwist) ,
+maxTwist_(maxTwist) ,
+maxTwistChange_(maxChange)
+{
+
+}
+
+// TODO sanity check these values
```
amiller27: Do we want to make this a compiler warning?  Are we not doing that anymore?

include/iarc7_motion/QuadTwistRequestLimiter.hpp
aftersomemath: Forgot

src/LowLevelMotionController.cpp
aftersomemath: Twist or a Pose, haven't decided yet.

src/LowLevelMotionController.cpp
aftersomemath: +1

src/QuadTwistRequestLimiter.cpp
aftersomemath: Not doing that anymore, they aren't being fixed and it clutters the build output unnecessarily.

src/QuadTwistRequestLimiter.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+void QuadTwistRequestLimiter::velocityLimit(double& request, const double old, const double max, const ros::Duration& delta, char const * axis)
+{
+    double velocity = (request - old) / delta.toSec();
+
+    if(abs(velocity) > max)
+    {
+        // result = (sign of velocity) * max_velocity + start
+        double result = (((velocity > 0) - (velocity < 0)) * max) + old;
+        ROS_ERROR("%s rate limit reached, requested: %f allowed: %f", axis, result, old);
```
aftersomemath: It is just so that we can print out the error. However, since I have written our own max, min, and slew rate functions for this class specifically, I think passing the string is warranted. It could be debated that these functions should not be defined in this class or we should be using some standard library function though.

src/LowLevelMotionController.cpp
amiller27: Well it isn't really either, because it has force (which would technically be a wrench, I think) and orientation, which would be a quaternion

src/QuadTwistRequestLimiter.cpp
amiller27: OK

PR #9, opened 2016-12-07
Low level motion integration test

amiller27:


src/FeedForwardPid.cpp
```diff
@@ -6,27 +6,45 @@
...
 }
 
 double FeedForwardPid::update(double current_value, double time_delta)
 {
-
-    static double i_accumulator{0.0};
-    static double last_current_value{0.0};
-
+    if (time_delta == 0) {
+        time_delta = std::numeric_limits<double>::min();
```
aftersomemath: What is the intention of this if block?

src/FeedForwardPid.cpp
```diff
@@ -6,27 +6,45 @@
...
+      i_accumulator_max_(i_accumulator_max),
+      i_accumulator_min_(i_accumulator_min)
 {
     // Nothing to do
 }
 
 double FeedForwardPid::update(double current_value, double time_delta)
 {
-
-    static double i_accumulator{0.0};
```
aftersomemath: I don't see anywhere where i_accumulator or last_current_value are used outside of the update function, I believe they should remain static inside this function to limit their scope.

src/LowLevelMotionController.cpp
```diff
@@ -90,12 +90,13 @@ int main(int argc, char **argv)
 
     ROS_INFO("Low_Level_Motion_Control begin");
 
-    ros::NodeHandle nh("low_level_motion_controller");
+    ros::NodeHandle nh;
+    ros::NodeHandle param_nh ("low_level_motion_controller");
```
aftersomemath: I like this convention, let's continue to use it.

src/LowLevelMotionController.cpp
```diff
@@ -110,12 +111,18 @@ int main(int argc, char **argv)
     Twist min;
     Twist max;
     Twist max_rate;
-    getUavCommandParams(nh, min, max, max_rate);
+    getUavCommandParams(param_nh, min, max, max_rate);
     QuadTwistRequestLimiter limiter(min, max, max_rate);
 
-    while(ros::ok())
+    ros::Rate rate (100);
+    while (ros::ok() && ros::Time::now() == ros::Time(0)) {
```
aftersomemath: This waits for the simulated time to start correct?

src/QuadVelocityController.cpp
```diff
@@ -16,12 +16,14 @@ using namespace Iarc7Motion;
 
 QuadVelocityController::QuadVelocityController(ros::NodeHandle& nh, double thrust_pid[3], double pitch_pid[3], double roll_pid[3]) :
 nh_(nh),  
-throttle_pid_(thrust_pid[0], thrust_pid[1], thrust_pid[2]),
-pitch_pid_(pitch_pid[0], pitch_pid[1], pitch_pid[2]) ,
-roll_pid_(roll_pid[0], roll_pid[1], roll_pid[2]) ,
-yaw_pid_(0.0, 0.0, 0.0) ,
+throttle_pid_(thrust_pid[0], thrust_pid[1], thrust_pid[2], 50, -50),
```
aftersomemath: The i accumulator limit values should be loaded from rosparam.

src/QuadVelocityController.cpp
```diff
@@ -31,27 +33,33 @@ void QuadVelocityController::setTargetVelocity(geometry_msgs::Twist twist)
...
     double pitch_output    = pitch_pid_.update(velocities.x, time_delta);
-    double roll_output     = roll_pid_.update(velocities.y, time_delta);
+    double roll_output     = roll_pid_.update(-velocities.y, time_delta);
     double yaw_output      = yaw_pid_.update(velocities.y, time_delta);
 
     // For now publish, should send values to a hard limiter first
     iarc7_msgs::OrientationThrottleStamped uav_command;
-    uav_command.throttle = throttle_output;
+    uav_command.header.stamp = time;
+    uav_command.throttle = throttle_output + hover_throttle_;
```
aftersomemath: This is fine as a temporary hack until the feedforward story is done.

src/QuadVelocityController.cpp
```diff
@@ -31,27 +33,33 @@ void QuadVelocityController::setTargetVelocity(geometry_msgs::Twist twist)
...
+iarc7_msgs::OrientationThrottleStamped QuadVelocityController::update(const ros::Time& time)
 {
-    // Get the time delta
     geometry_msgs::Vector3 velocities;
     getVelocities(velocities);
-    double time_delta = 0.0;
+
+    // Get the time delta
+    if (last_time_ == ros::Time(0)) {
+        last_time_ = time;
```
aftersomemath: I think this should be restructured to:

`if(last_time !-= ros::Time(0))
{ // do normal pid
}
else
{
// return zerod out uav_command
}`

What we do know will automatically have a throttle value which I feel is slightly unsafe because we don't have a valid delta yet.

src/FeedForwardPid.cpp
amiller27: I believe static locals are shared between instances (I tested it), and we don't want these to be shared

src/LowLevelMotionController.cpp
amiller27: I agree, but it's not really a matter of convention.  Doing it the other way requires that you use global names for topics which aren't in the node's namespace, which prevents you from running the node inside another namespace

src/LowLevelMotionController.cpp
amiller27: Yes

src/QuadVelocityController.cpp
amiller27: Agreed, I'll fix this

src/FeedForwardPid.cpp
aftersomemath: I meant to check that myself before leaving this comment. Sounds good!

src/LowLevelMotionController.cpp
aftersomemath: Ok

src/QuadVelocityController.cpp
amiller27: Agreed

src/QuadVelocityController.cpp
amiller27: I think the PID controller should be able to return a value on the first call, using only the P term.  I can think of a couple ways to do that: we could change the PID controller to take a time instead of a delta, just keep track of whether the PID controller has been updated before, or do what I have here.  The upside to what I have here is that it works if the PID controller gets called with the same time after it's been called the first time

aftersomemath: I can't think of a use case where we need a PID controller to not implement the full PID model its designed for on every iteration. Its also opening us up for accidental behaviour if its not explicitly called out in the PID class.

In most applications you'll never notice that only the p term is returned on the first output.

src/FeedForwardPid.cpp
aftersomemath: OK I think a comment here might be useful in the future

src/QuadVelocityController.cpp
aftersomemath: A delta of 0 is a valid input to the PID.

The PID is always initialized upon object creation this was by design.

I would argue that warnings should always be sent when receiving a time delta of zero in a PID controller.

However because this class is a quad controller and there is no reason it shouldn't wait to be fully initialized before sending flight commands.

src/FeedForwardPid.cpp
aftersomemath: Also since we are taking the time I think we should properly handle this. Not just use min. It works for this applications output scale but would not always be a good thing.

amiller27: Really it should be some multiple of the granularity of whatever clock we're using, but that's hard to get.  We could just guess on the granularity

amiller27: Really min isn't actually good for any application, because it's way smaller than the clock granularity on any system unless your units of time are ridiculous

amiller27: Like if you measured time in units of the age of the universe and your clock had a granularity of the planck time it would still be too small

aftersomemath: Hmmm. The real problem is that a if the input value changes and the time delta is zero than we have a step function on our input value which is impossible physically.

I think we should be erroring at this point and the caller should handle the error.

Otherwise the value we put out should be the result of discarding the results of the previous addition to the I_accumulator recalculating p. And recalculating d using the last last setpoint.

aftersomemath: I'd like to avoid being dependent on our time librarys implantation properties that don't have much to do with theoretical time.

amiller27: I disagree.  It's not impossible physically, because the granularity on your clock is finite, so if the value changes before your clock updates then you get the input changing with a time delta of 0.  In that case, we should really assume the time delta is approximately the clock granularity.  It would make sense to log that this is happening though, probably as a warning.  I would also like to avoid being dependent on our time library's implementation, so what we could do is have a parameter passed in for the granularity of the clock.

aftersomemath: A p controller can do what you want to do here. A pid controller can not always accurately do it. There are control cases where what you want to do will not be accurate particularly if your timestep granularity is coarse.

I think the software is making a false promise if it allows for control values to be updated faster than a somewhat significant multiple of the clocks granularity. (In the case where the update rate is variable)

In the case that the update rate is fixed to a multiple of the clocks granularity. Input values should be sampled as close to the entry of the PID loop as possible.

In either case not doing so opens up your PID loop possible spikes in control output.

amiller27: You could deal with spikes by using the max possible time for the derivative and 0 for the integral, but I'm thinking that just making it illegal to pass in 0 for the delta would be better practice

aftersomemath: I agree that that will be easier and clearer to the end user. Return a NaN?

amiller27: I think I want to return a bool because then we can turn on compiler warnings if the return value is ignored

amiller27: I like the idea of returning a NaN, but I think it makes it easy for users not to check for failure

amiller27: I disagree. Warnings are for things that are probably fine but are good to know about. Errors are for things that we might need to fail on

src/FeedForwardPid.cpp
```diff
@@ -17,20 +19,47 @@ FeedForwardPid::FeedForwardPid(double p_gain, double i_gain, double d_gain,
...
 {
     // Nothing to do
 }
 
-double FeedForwardPid::update(double current_value, double time_delta)
+bool FeedForwardPid::update(double current_value, double time_delta, double& response)
 {
-    if (time_delta == 0) {
-        time_delta = std::numeric_limits<double>::min();
+    if (!initialized_ && std::isfinite(current_value)) {
```
aftersomemath: I still don't agree with this, but for the sake of time it's ok. Additionally, I think you should be including the i_term in this since it can be calculated on the first run as well.

src/QuadVelocityController.cpp
```diff
@@ -49,16 +50,40 @@ iarc7_msgs::OrientationThrottleStamped QuadVelocityController::update(const ros:
...
-    double pitch_output    = pitch_pid_.update(velocities.x, time_delta);
-    double roll_output     = roll_pid_.update(-velocities.y, time_delta);
-    double yaw_output      = yaw_pid_.update(velocities.y, time_delta);
+    double throttle_output;
+    double pitch_output;
+    double roll_output;
+    double yaw_output;
+
+    bool success = throttle_pid_.update(velocities.z, time_delta, throttle_output);
+    if (!success) {
```
aftersomemath: http://wiki.ros.org/rosconsole section on conditionals. Can we use these here?

amiller27: Yep, I actually didn't know about those

amiller27: Actually I don't want to do that, because in every case I want to return differently if I display the warning.  I could do it in FeedForwardPid::update, but I think it would make the code less readable

src/FeedForwardPid.cpp
```diff
@@ -6,29 +6,81 @@
...
+      last_current_value_(0.0),
+      setpoint_(0.0),
+      i_accumulator_max_(i_accumulator_max),
+      i_accumulator_min_(i_accumulator_min)
 {
     // Nothing to do
 }
 
-double FeedForwardPid::update(double current_value, double time_delta)
+bool FeedForwardPid::update(double current_value, double time_delta, double& response)
```
aftersomemath: This function allows for zero to be time_delta on the first run and not on the second. This is inconsistent behavior.

amiller27: That was the intention, because time_delta is ignored on the first call.  I couldn't think of a simpler way to do it

aftersomemath: I think the problem is that one function has two purposes that are related to the same action.

To make the code cleaner this function could be broken into another simpler one and called from here.

amiller27: I think the cleanest way would be to call this with a time instead of a delta, and then it's called with the same information every time.  The only problem with that is if you were trying to keep this class independent of ROS

src/QuadVelocityController.cpp
aftersomemath: This particular function does not need to perform an early return to work properly, and early returns make code less readable. I think it would be ok to use the conditional message and follow through with calling the rest of the PID updates.

amiller27: So if an update fails, we want to still call the other updates and return the failure at the end?

aftersomemath: Yes, there's a small chance it provides more debugging information come to think of it.

PR #11, opened 2016-12-21
Support simultaneous yaw and translation

aftersomemath:
Support simultaneous yaw and translation, refactor QuadVelocity controller to use twists.

The waypoint tester was expanded to test this functionality.

Currently yaw calculations are done using the angle around the z axis. Using Quanternions in the future is an idea for a future rewrite.

### DIFF
--- a/launch/sim_low_level_motion.launch
+++ b/launch/sim_low_level_motion.launch
@@ -1,5 +1,5 @@
 <launch>
-    <node name="low_level_motion_controller" pkg="iarc7_motion" type="low_level_motion_controller">
+    <node name="low_level_motion_controller" pkg="iarc7_motion" type="low_level_motion_controller" output="screen">
         <rosparam command="load" file="$(find iarc7_motion)/param/sim_low_level_motion.yaml" />
     </node>
 </launch>
--- a/launch/sim_low_level_motion.launch
+++ b/launch/sim_low_level_motion.launch
@@ -1,5 +1,5 @@
 <launch>
-    <node name="low_level_motion_controller" pkg="iarc7_motion" type="low_level_motion_controller" output="screen">
+    <node name="low_level_motion_controller" pkg="iarc7_motion" type="low_level_motion_controller">
         <rosparam command="load" file="$(find iarc7_motion)/param/sim_low_level_motion.yaml" />
     </node>
 </launch>
--- a/launch/sim_low_level_motion.launch
+++ b/launch/sim_low_level_motion.launch
@@ -1,5 +1,5 @@
 <launch>
     <node name="low_level_motion_controller" pkg="iarc7_motion" type="low_level_motion_controller">
-        <rosparam command="load" file="$(find iarc7_motion)/param/sim_low_level_motion.yaml" />
+        <rosparam command="load" file="$(find iarc7_motion)/param/sim_low_level_motion.yaml"/>
     </node>
 </launch>
--- a/scripts/waypoints_test.py
+++ b/scripts/waypoints_test.py
@@ -58,7 +58,7 @@ if __name__ == '__main__':
         # Get the yaw (z axis) rotation from the quanternion
         current_yaw = tf.transformations.euler_from_quaternion(rot)[2]
         
-        # Transform current yaw to be between 0 and 2pi
+        # Transform current yaw to be between 0 and 2pi because the points are encoded from 0 to 2pi
         if current_yaw < 0:
             current_yaw = (2.0 * math.pi) + current_yaw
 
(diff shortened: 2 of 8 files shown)


launch/sim_low_level_motion.launch
```diff
@@ -1,5 +1,5 @@
 <launch>
-    <node name="low_level_motion_controller" pkg="iarc7_motion" type="low_level_motion_controller">
+    <node name="low_level_motion_controller" pkg="iarc7_motion" type="low_level_motion_controller" output="screen">
```
amiller27: I really don't like having this in the launch file, it should be passed on the command line of you want the output displayed

scripts/waypoints_test.py
```diff
@@ -49,6 +54,32 @@ def constrain(x, l, h):
             velocity.twist.linear.y = constrain((target[1] - trans[1]) * kP, -max_vel, max_vel)
         if abs(target[2] - trans[2]) >= 0.02:
             velocity.twist.linear.z = target[2] - trans[2]
+        
+        # Get the yaw (z axis) rotation from the quanternion
+        ysqr = rot[1] * rot[1];
```
amiller27: Actually this whole thing should probably be done with the TF library

src/QuadVelocityController.cpp
```diff
@@ -122,19 +128,46 @@ bool QuadVelocityController::getVelocities(geometry_msgs::Vector3& return_veloci
                 velocities_valid = false;
             }
 
-            // Get the transforms with the stamps for readability
+            // Get the transforms without the stamps for readability
             geometry_msgs::Transform& transform = transformStamped.transform;
             geometry_msgs::Transform& oldTransform = last_transform_stamped_.transform;
 
+            // Get the yaw (z axis) rotation from the quanternion
+            double ysqr = transform.rotation.y * transform.rotation.y;
```
amiller27: We should be doing this using ROS

src/QuadVelocityController.cpp
```diff
@@ -122,19 +128,46 @@ bool QuadVelocityController::getVelocities(geometry_msgs::Vector3& return_veloci
...
+
+            return_velocities.linear.z = (transform.translation.z - oldTransform.translation.z) / delta;
+
+            double yaw_difference = current_yaw - last_yaw_;
+
+            // The next two if statements handle if the yaw_difference is large due to gimbal lock
+            // Assumes that we don't have a yaw_difference more than 2pi. We won't since we get all the angles
+            // from an atan2 which only return -pi to pi
+            // Also assumes that we won't have a rotation difference more than pi
+            if(yaw_difference > 3.14159)
```
amiller27: We should use a builtin constant here

launch/sim_low_level_motion.launch
aftersomemath: Agreed that was leftover from debugging.

src/QuadVelocityController.cpp
aftersomemath: C++ doesn't have a built-in constant in the spec: http://stackoverflow.com/questions/1727881/how-to-use-the-pi-constant-in-c

It's also defined in boost.

However it's available in gcc easily these day's it's just not standard.

scripts/waypoints_test.py
aftersomemath: I don't think that's a huge deal since this is test code and it works as is.

aftersomemath: However I will change it here, as it will be more readable and there are no speed requirements.

src/QuadVelocityController.cpp
aftersomemath: I disagree in this case. Doing it via ROS requires constructing a Quaternion object, then constructing a Matrix3x3 object with the Quaternion, and then calling a method on the matrix to get the ypr. Which in the end means we constructed two objects to calculate three angles when we only needed to do a little math to get one angle.

src/QuadVelocityController.cpp
```diff
@@ -33,16 +35,20 @@ ran_once_(false)
...
 {
     throttle_pid_.setSetpoint(twist.linear.z);
-    pitch_pid_.setSetpoint(twist.linear.x);
-    roll_pid_.setSetpoint(-twist.linear.y);
     yaw_pid_.setSetpoint(twist.angular.z);
+
+    // Pitch and roll velocities are transformed to vectors that pitch and roll can respond to
+    pitch_pid_.setSetpoint(twist.linear.x * std::cos(last_yaw_) + twist.linear.y * std::sin(last_yaw_));
+
+    // Invert roll because a positive y velocity means a negative roll by the right hand rule
```
aftersomemath: I think this should be fixed when feed forward is implemented. The PID loop should take care of the fact you use negative feedback to get a positive velocity.

aftersomemath: Let's move this out of the code review since it's not really related.

src/QuadVelocityController.cpp
amiller27: This is only running at like 100Hz, if the jetson can't afford to create 200 extra objects per second we have other problems

amiller27: I guess I'm ok if we leave it though

PR #13, opened 2016-12-23
Low level motion safety response

aftersomemath:
Use the new features from iarc7_safety to create safety responses for low level motion and the waypoints test.

### DIFF
--- a/src/LowLevelMotionController.cpp
+++ b/src/LowLevelMotionController.cpp
@@ -12,6 +12,8 @@
 #include "iarc7_motion/QuadVelocityController.hpp"
 #include "iarc7_motion/QuadTwistRequestLimiter.hpp"
 
+#include "iarc7_safety/SafetyClient.hpp"
+
 #include "geometry_msgs/Twist.h"
 #include "geometry_msgs/TwistStamped.h"
 
@@ -106,6 +108,10 @@ int main(int argc, char **argv)
     ros::NodeHandle nh;
     ros::NodeHandle param_nh ("low_level_motion_controller");
 
+    ROS_INFO("low_level_motion: Attempting to form safety bond");
+    Iarc7Safety::SafetyClient safety_client(nh, "low_level_motion_bond");
+    ROS_ASSERT_MSG(safety_client.formBond(), "low_level_motion: Could not form bond with safety client");
+
     double throttle_pid[5];
     double pitch_pid[5];
     double roll_pid[5];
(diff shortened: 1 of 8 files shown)


src/LowLevelMotionController.cpp
```diff
@@ -136,6 +142,16 @@ int main(int argc, char **argv)
...
     {
+        // Check the safety client before updating anything
+        ROS_ASSERT_MSG(!safety_client.isFatalActive(), "low_level_motion: fatal event from safety");
+
+        if(safety_client.isSafetyActive())
+        {
+            // Override whatever the Acceleration Planner wants to do and attempt to land
+            geometry_msgs::Twist twist;
+            twist.linear.z = -0.2; // Try to descend
+            quadController.setTargetVelocity(twist);
```
amiller27: Don't we also need to tell the quadController to not accept new requested velocities?

aftersomemath: No because the only time the quad controller is updated is after this call. Thus this will always override the velocity set by the acceleration planner.

I don't think the QuadVelocity controller should be safety aware (single responsibility) and I don't want it to have a way to stop accepting velocities in case the concept of safety is ever expanded to include recovery.

amiller27: I see, because the accelerationplanner is only updated in a callback.  It seems like this would be very easy to break in the future because it's dependent on ordering of statements that don't have an obvious need to be ordered that way.  I agree that the quadvelocitycontroller shouldn't be safety aware, but it might be useful to have a function in the quadvelocitycontroller that disables/enables the acceleration planner

aftersomemath: I agree. I dislike the idea of modifying the quadvelocity controller to be aware of it's input source that's far outside of the responsibility of the class.

After thinking about it a bit, I realized that the structure of the acceleration planner doesn't match the rest of the program. It shouldn't be responsible for executing it's update functions periodically like it is. (This also means the correct velocities are not always being sent to the velocity controller because of the difference in when the velocities are generated and the velocity controller updating). I will refactor it to be updated from the main program loop, this will double in making the intentions of the safety functionality clear and refrain from doing unnecessary, inaccurate calculations.

amiller27: That would be great

aftersomemath: Fixed

PR #15, opened 2016-12-31
Fix blocking call to lookupTransform

amiller27:
The QuadVelocityController::getVelocities function now should always
return a valid velocity unless something goes wrong, no more
first-velocity-invalid stuff.  Also there is now a giant loop where the
blocking call to lookupTransform used to be, which makes sure events are
handled while waiting and cases like waiting on a transform when newer
ones are available aren't a problem.

### DIFF
--- a/src/QuadVelocityController.cpp
+++ b/src/QuadVelocityController.cpp
@@ -6,6 +6,8 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#include <sstream>
+
 // Associated header
 #include "iarc7_motion/QuadVelocityController.hpp"
 
@@ -26,6 +28,7 @@ yaw_pid_(yaw_pid[0], yaw_pid[1], yaw_pid[2], yaw_pid[3], yaw_pid[4]),
 hover_throttle_(58.0),
 last_transform_stamped_(),
 last_yaw_(0.0),
+have_last_velocity_stamped_(false),
 ran_once_(false)
 {
 
@@ -49,7 +52,11 @@ bool QuadVelocityController::update(const ros::Time& time,
                                     iarc7_msgs::OrientationThrottleStamped& uav_command)
 {
     geometry_msgs::Twist velocities;
-    getVelocities(velocities);
+    bool success = getVelocities(velocities);
+    if (!success) {
+        ROS_WARN("Failed to get current velocities in QuadVelocityController::update");
+        return false;
+    }
 
     // Update all the PID loops
     double throttle_output;
@@ -57,7 +64,7 @@ bool QuadVelocityController::update(const ros::Time& time,
     double roll_output;
     double yaw_output;
 
-    bool success = throttle_pid_.update(velocities.linear.z, time, throttle_output);
+    success = throttle_pid_.update(velocities.linear.z, time, throttle_output);
     if (!success) {
         ROS_WARN("Throttle PID update failed in QuadVelocityController::update");
         return false;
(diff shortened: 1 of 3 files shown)


src/QuadVelocityController.cpp
```diff
@@ -109,13 +116,52 @@ bool QuadVelocityController::getVelocities(geometry_msgs::Twist& return_velociti
...
     // Check if we already have a velocity at this time
-    if (time == last_velocity_stamped_.header.stamp) {
+    if (have_last_velocity_stamped_ && time == last_velocity_stamped_.header.stamp) {
         return_velocities = last_velocity_stamped_.twist;
         return true;
     }
 
+    // We don't have a velocity at this time yet, so make sure we don't fetch
+    // a transform at the same time as the last one
+    while (ran_once_ && time == last_transform_stamped_.header.stamp) {
```
aftersomemath: Why is this more desirable than returning from the function?

src/QuadVelocityController.cpp
```diff
@@ -167,17 +213,21 @@ bool QuadVelocityController::getVelocities(geometry_msgs::Twist& return_velociti
             return_velocities.angular.z = yaw_difference / delta;
 
             // Store this as the last valid velocity
+            have_last_velocity_stamped_ = true;
             last_velocity_stamped_.twist = return_velocities;
             last_velocity_stamped_.header.stamp = time;
+
+            last_transform_stamped_ = transformStamped;
```
aftersomemath: These lines are duplicated in both the if and the else statement. Can they be moved out of the if statement to avoid code duplication? I don't think it improves clarity in this case to have them duplicated in the if and the else.

src/QuadVelocityController.cpp
```diff
@@ -109,13 +116,52 @@ bool QuadVelocityController::getVelocities(geometry_msgs::Twist& return_velociti
...
 
+    // We don't have a velocity at this time yet, so make sure we don't fetch
+    // a transform at the same time as the last one
+    while (ran_once_ && time == last_transform_stamped_.header.stamp) {
+        time = ros::Time::now();
+    }
+
     try{
-        transformStamped = tfBuffer_.lookupTransform("map", "quad", time, ros::Duration(MAX_TRANSFORM_WAIT_SECONDS));
+        bool found_transform = false;
```
aftersomemath: This block of code seems like a good candidate for a new private class method. I think it's kinda messy here.

src/QuadVelocityController.cpp
amiller27: Because the purpose of the function seems to be to block until a new velocity is ready

src/QuadVelocityController.cpp
amiller27: No, because they have to run before the following function call

src/QuadVelocityController.cpp
amiller27: Could be, I'll look at refactoring

src/QuadVelocityController.cpp
aftersomemath: The purpose was to update the pid controllers with info from the current timestamp. However it can be changed to what you said.

Should this blocking loop call ros::OK and spinonce? Could simulated time mess with it if you don't call those functions?

src/QuadVelocityController.cpp
aftersomemath: Ok

PR #20, opened 2017-01-24
Comment low level motion

aftersomemath:
Fix a logic error in the acceleration planner

Simplifying the transform and velocity getting functions.

### DIFF
--- a/src/QuadVelocityController.cpp
+++ b/src/QuadVelocityController.cpp
@@ -220,23 +220,21 @@ bool QuadVelocityController::waitForNewVelocities(geometry_msgs::Twist& return_v
         last_transform_stamped_ = transformStamped;
         return waitForNewVelocities(return_velocities);
     }
-    // If we have ran before make sure the difference between transforms isn't too much
-    else
+
+    // Make sure the difference between transforms isn't too much
+    // Calculate the lastest time between transforms that is allowed.
+    ros::Time latest_time_allowed = last_transform_stamped_.header.stamp
+                + ros::Duration(MAX_TRANSFORM_DIFFERENCE_SECONDS);
+    // Make the sure the transform isn't didn't come after the last time
+    // that was acceptable.
+    if (transformStamped.header.stamp > latest_time_allowed)
     {
-        // Calculate the lastest time between transforms that is allowed.
-        ros::Time latest_time_allowed = last_transform_stamped_.header.stamp
-                    + ros::Duration(MAX_TRANSFORM_DIFFERENCE_SECONDS);
-        // Make the sure the transform isn't didn't come after the last time
-        // that was acceptable.
-        if (transformStamped.header.stamp > latest_time_allowed)
-        {
-            ROS_ERROR_STREAM(
-                "First available transform came after latest allowed time"
-                << " (transform time: " << transformStamped.header.stamp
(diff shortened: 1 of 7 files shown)


src/QuadVelocityController.cpp
```diff
@@ -129,77 +128,38 @@ bool QuadVelocityController::update(const ros::Time& time,
...
-bool QuadVelocityController::getTransformAfterTime(
-        const ros::Time& time,
-        geometry_msgs::TransformStamped& transform,
-        const ros::Time& latest_time_allowed) {
+bool QuadVelocityController::waitForTransform(geometry_msgs::TransformStamped& transform) {
 
     // Catch TransformException exceptions
     try
     {
+        ros::Time time_at_start = ros::Time::now();
```
amiller27: I really think this should be the timestamp of the last transform we processed.  Otherwise we risk outputting at lower frequency than the incoming data

amiller27: Also if we change this we might want the timeout at line 142 to be measured differently

src/QuadVelocityController.cpp
```diff
@@ -213,59 +173,63 @@ bool QuadVelocityController::getTransformAfterTime(
...
+    // Update all intermediate values and make a recursive call to waitForNewVelocities
+    // (meaning another transform will be collected) and a valid velocity can be calculated.
+    if(!wait_for_velocities_ran_once_)
+    {
+        wait_for_velocities_ran_once_ = true;
+        last_transform_stamped_ = transformStamped;
+        return waitForNewVelocities(return_velocities);
+    }
+    // If we have ran before make sure the difference between transforms isn't too much
+    else
```
amiller27: Either the entire body of the function after this should be inside this `else` block, or none of it should.  I vote none of it, if you put all of it, then for consistency, we should put everything after line 229 inside another `else` on that `if` condition.

src/QuadVelocityController.cpp
```diff
@@ -213,59 +173,63 @@ bool QuadVelocityController::getTransformAfterTime(
...
-            && ran_once_
-            && time == last_transform_stamped_.header.stamp) {
-        ros::spinOnce();
-    }
-
-    // Calculate the lastest time between transforms that is allowed.
-    ros::Time latest_time_allowed;
-    if (ran_once_)
+    // Make sure that the current time is more than the last transform stamped time
+    while (wait_for_velocities_ran_once_
```
amiller27: I don't really think we should be doing this because it prevents us from processing future dated transforms if they're available.

src/QuadVelocityController.cpp
aftersomemath: Agreed. There is no problem with doing this because we get the latest transform anyway. So if two transforms come in after the last processed transform the latest will still be used and we won't miss transforms if running too slow.

src/QuadVelocityController.cpp
aftersomemath: Why should we be using a future dated velocity?

src/QuadVelocityController.cpp
aftersomemath: Fixed

src/QuadVelocityController.cpp
aftersomemath: Fixed

PR #31, opened 2017-02-16
Add takeoff task

aftersomemath:
There is a lot here.

Motion planner class now handles all command requests from tasks.
Takeoff task has a bunch of states to cycle through before allowing takeoff.
Updated test task to reflect changes in the motion planner structure.
Renamed test_motion_planner.py to test_action_server.py
Created test_motion_planner.py

### DIFF
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -1,4 +1,4 @@
 takeoff_velocity        : 2.0
 takeoff_height_tolerance: 0.1
-delay_before_takeoff: 2.0
+delay_before_takeoff    : 2.0
 max_takeoff_start_height: 0.5
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -1,0 +1,3 @@
+takeoff_velocity        : 2.0
+takeoff_height_tolerance: 0.1
+delay_before_takeoff: 2.0
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -1,6 +1,13 @@
+# Velocity with which to take off with
 takeoff_velocity        : 2.0
+# If we get to this height less than the requested take off height
+# it is counted as take off occuring
 takeoff_height_tolerance: 0.1
+# Delay between arming and taking off
 delay_before_takeoff    : 2.0
+# Maximum height considered valid before taking off
 max_takeoff_start_height: 0.3
+# Velocity with which to land with
 land_velocity           : -1.0
+# Height below which landing is considered finish
 land_height_tolerance   : 0.3
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -1,6 +1,6 @@
 takeoff_velocity        : 2.0
 takeoff_height_tolerance: 0.1
 delay_before_takeoff    : 2.0
-max_takeoff_start_height: 0.5
+max_takeoff_start_height: 0.3
 land_velocity           : -1.0
-land_height_tolerance   : 0.5
+land_height_tolerance   : 0.3
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -1,3 +1,4 @@
 takeoff_velocity        : 2.0
 takeoff_height_tolerance: 0.1
 delay_before_takeoff: 2.0
+max_takeoff_start_height: 0.5
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -2,3 +2,5 @@ takeoff_velocity        : 2.0
 takeoff_height_tolerance: 0.1
 delay_before_takeoff    : 2.0
 max_takeoff_start_height: 0.5
+land_velocity           : -1.0
+land_height_tolerance   : 0.5
--- a/scripts/iarc_task_action_server.py
+++ b/scripts/iarc_task_action_server.py
@@ -37,16 +37,17 @@ class IarcTaskActionServer:
             task_request = goal.get_goal()
 
             try:
-                actionvalues_dict = {'takeoff_height': task_request.takeoff_height ,
-                                     'preempt': task_request.preempt , 
-                                     'movement_type': task_request.movement_type}
+                new_task_type = self._task_dict[task_request.movement_type]
             except KeyError as e:
                 rospy.logerr("Goal has invalid movement_type: %s", task_request.movement_type)
                 goal.set_rejected()
                 return
 
             try:
-                new_task = self._task_dict[task_request.movement_type](actionvalues_dict)
+                actionvalues_dict = {'takeoff_height': task_request.takeoff_height ,
+                                     'preempt': task_request.preempt , 
+                                     'movement_type': task_request.movement_type}
+                new_task = new_task_type(actionvalues_dict)
             except Exception as e:
                 rospy.logerr("Could not construct task: %s", task_request.movement_type)
                 rospy.logerr(str(e))
--- a/scripts/iarc_task_action_server.py
+++ b/scripts/iarc_task_action_server.py
@@ -3,6 +3,7 @@ import threading
 import rospy
 import actionlib
 from takeoff_task import TakeoffTask
+from land_task import LandTask
 from test_task import TestTask
 from iarc7_motion.msg import QuadMoveAction, QuadMoveResult
 
@@ -25,6 +26,7 @@ class IarcTaskActionServer:
         self._action_server.start()
 
         self._task_dict = {'takeoff': TakeoffTask,
+                           'land': LandTask,
                            'test_task': TestTask}
 
     # Private method
--- a/scripts/iarc_task_action_server.py
+++ b/scripts/iarc_task_action_server.py
@@ -38,9 +38,16 @@ class IarcTaskActionServer:
                 actionvalues_dict = {'takeoff_height': task_request.takeoff_height ,
                                      'preempt': task_request.preempt , 
                                      'movement_type': task_request.movement_type}
-                new_task = self._task_dict[task_request.movement_type](actionvalues_dict)
             except KeyError as e:
                 rospy.logerr("Goal has invalid movement_type: %s", task_request.movement_type)
+                goal.set_rejected()
+                return
+
+            try:
+                new_task = self._task_dict[task_request.movement_type](actionvalues_dict)
+            except Exception as e:
+                rospy.logerr("Could not construct task: %s", task_request.movement_type)
+                rospy.logerr(str(e))
                 goal.set_rejected()
                 return
 
--- a/scripts/land_task.py
+++ b/scripts/land_task.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python
+import rospy
+from abstract_task import AbstractTask
+from task_state import TaskState
+from geometry_msgs.msg import TwistStamped
+
+import rospy
+import tf2_ros
+
+import math
+
+from iarc7_msgs.msg import FlightControllerStatus
+from geometry_msgs.msg import TwistStamped
+
+class LandTaskState:
+    init = 0
+    land = 1
+    disarm = 2
+
+class LandTask(AbstractTask):
+
+    def __init__(self, actionvalues_dict):
+        self.tf_buffer = tf2_ros.Buffer()
+        self.tf_listener = tf2_ros.TransformListener(self.tf_buffer)
+        self.canceled = False;
+
+        try:
+            self.LAND_VELOCITY = rospy.get_param('~land_velocity')
+            self.LAND_HEIGHT_TOLERANCE = rospy.get_param('~land_height_tolerance')
+        except KeyError as e:
+            rospy.logerr('Could not lookup a parameter for takeoff task')
+            raise
+        
+        self.fc_status = None
+        self.fc_status_sub = rospy.Subscriber('fc_status', FlightControllerStatus, self._receive_fc_status)
+        self.state = LandTaskState.init
+        self.disarm_request_success = False
+
+    def _receive_fc_status(self, data):
+        self.fc_status = data
+
+    def disarm_callback(self, data):
+        self.disarm_request_success = data.success
+
+    def get_desired_command(self):
+        if self.canceled:
+            return (TaskState.canceled)
+
+        if self.state == LandTaskState.init:
+            if self.fc_status != None:
+                # Check that auto pilot is enabled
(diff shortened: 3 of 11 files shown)


scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def get_desired_command(self):
+        if self.canceled:
+            return (TaskState.canceled)
+
+        if self.state == LandTaskState.init:
+            if self.fc_status != None:
+                # Check that auto pilot is enabled
+                if not self.fc_status.auto_pilot:
+                    return (TaskState.failed, 'flight controller not allowing auto pilot')
```
amiller27: These return states should really be classes, with some of them having data members (like the failure message).  I vote we make a trello card to refactor with that feature

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+                # All is good change state to land
+                else:
+                    self.state = LandTaskState.land
+            else:
+                return (TaskState.running, 'velocity', TwistStamped())
+
+        # Enter the takeoff phase
+        if self.state == LandTaskState.land:
+            try:
+                transStamped = self.tf_buffer.lookup_transform('map', 'quad', rospy.Time(0))
```
amiller27: This should be the only line in the try block because it's the only line that throws the exception you're looking for

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+                # Check if we reached the target height
+                if(transStamped.transform.translation.z < self.LAND_HEIGHT_TOLERANCE):
+                    self.state = LandTaskState.disarm
+
+                velocity = TwistStamped()
```
amiller27: This should only happen if we didn't switch into the disarm state

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+                velocity.header.frame_id = 'level_quad'
+                velocity.header.stamp = rospy.Time.now()
+                velocity.twist.linear.z = self.LAND_VELOCITY
+                return (TaskState.running, 'velocity', velocity)
+
+            except (tf2_ros.LookupException, tf2_ros.ConnectivityException, tf2_ros.ExtrapolationException) as ex:
+                rospy.logerr('Takeofftask: Exception when looking up transform')
```
amiller27: Wrong error message, also two lines below here

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+                rospy.logerr(ex.message)
+                return (TaskState.aborted, 'Exception when looking up transform during takeoff')
+
+        # Enter the disarming request stage
+        if self.state == LandTaskState.disarm:
+            # Check if disarm succeeded
+            if self.disarm_request_success:
+                return (TaskState.done, 'velocity', TwistStamped())
```
amiller27: Why isn't this action a nop?  We shouldn't request a velocity anyway when the motors aren't armed

scripts/takeoff_task.py
```diff
@@ -4,14 +4,116 @@
 from task_state import TaskState
```
amiller27: Most comments from land_task.py are also relevant here

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def disarm_callback(self, data):
+        self.disarm_request_success = data.success
+
+    def get_desired_command(self):
+        if self.canceled:
+            return (TaskState.canceled)
+
+        if self.state == LandTaskState.init:
+            if self.fc_status != None:
```
amiller27: Use `is not None` instead of `!= None`

param/motion_planner.yaml
```diff
@@ -0,0 +1,6 @@
+takeoff_velocity        : 2.0
```
aftersomemath: Fixed

scripts/iarc_task_action_server.py
```diff
@@ -38,12 +40,19 @@ def _new_goal(self, goal):
...
                                      'movement_type': task_request.movement_type}
-                new_task = self._task_dict[task_request.movement_type](actionvalues_dict)
             except KeyError as e:
                 rospy.logerr("Goal has invalid movement_type: %s", task_request.movement_type)
                 goal.set_rejected()
                 return
 
+            try:
+                new_task = self._task_dict[task_request.movement_type](actionvalues_dict)
+            except Exception as e:
```
aftersomemath: Fixed

scripts/test_action_server.py
```diff
@@ -0,0 +1,77 @@
+#! /usr/bin/env python
```
aftersomemath: I'm not going to do that right now.

scripts/motion_planner.py
```diff
@@ -9,47 +13,92 @@ class MotionPlanner:
     def __init__(self, _action_server):
         self._action_server = _action_server
         self.task = None
```
aftersomemath: Fixed

scripts/motion_planner.py
```diff
@@ -9,47 +13,92 @@ class MotionPlanner:
...
+        self.velocity_pub = rospy.Publisher('movement_velocity_targets', TwistStampedArrayStamped, queue_size=0)
+        self.arm_service = rospy.ServiceProxy('uav_arm', SetBool)
```
aftersomemath: Will appear in upcoming PR for adding a safety response, sorry for the garble

scripts/motion_planner.py
```diff
@@ -9,47 +13,92 @@ class MotionPlanner:
...
-    def get_velocity_command(self):
+    def run(self):
+        rate = rospy.Rate(10)
+        while not rospy.is_shutdown():
+            # Tuples could have different lengths so just get the whole tuple
+            task_command = self._get_task_command()
+            if(task_command[0] == 'velocity'):
+                self._publish_twist(task_command[1])
+            elif(task_command[0] == 'arm'):
+                task_command[1](self._use_arm_service(True))
```
aftersomemath: Fixed

scripts/motion_planner.py
```diff
@@ -9,47 +13,92 @@ class MotionPlanner:
...
+            elif(task_command[0] == 'disarm'):
+                task_command[1](self._use_arm_service(False))
+            elif(task_command[0] == 'nop'):
+                self._publish_twist(TwistStamped())
```
aftersomemath: Fixed

scripts/motion_planner.py
```diff
@@ -9,47 +13,92 @@ class MotionPlanner:
...
+            else:
+                rospy.logerr('Unkown command request: %s, noping', task_command[0])
+                self._publish_twist(TwistStamped())
```
aftersomemath: Fixed

scripts/motion_planner.py
```diff
@@ -9,47 +13,92 @@ class MotionPlanner:
...
+        velocity_msg.data = [twist]
+        self.velocity_pub.publish(velocity_msg)
+
+    def _get_task_command(self):
         if (self.task is None) and self._action_server.has_new_task():
             self.task = self._action_server.get_new_task()
 
         if self.task:
             if self._action_server.is_canceled():
                 self.task.cancel()
```
aftersomemath: Exceptions from the action server? Im ok with letting the motion planner crash at that point.

scripts/motion_planner.py
```diff
@@ -9,47 +13,92 @@ class MotionPlanner:
...
 if __name__ == '__main__':
     rospy.init_node('motion_planner')
-
     action_server = IarcTaskActionServer()
 
     motion_planner = MotionPlanner(action_server)
+    try:
+        motion_planner.run()
+    except Exception, e:
+        rospy.logfatal("Error in motion planner get velocity command.")
```
aftersomemath: Fixed

scripts/motion_planner.py
```diff
@@ -9,47 +13,92 @@ class MotionPlanner:
...
+        rospy.logfatal(str(e))
+        rospy.signal_shutdown("Motion Planner shutdown")
```
aftersomemath: I do not believe that will allow me to raise the exception again.

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python
+import rospy
+from abstract_task import AbstractTask
+from task_state import TaskState
+from geometry_msgs.msg import TwistStamped
+
+import rospy
```
aftersomemath: Fixed

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+import tf2_ros
+
+import math
+
+from iarc7_msgs.msg import FlightControllerStatus
+from geometry_msgs.msg import TwistStamped
```
aftersomemath: Fixed

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+
+class LandTaskState:
+    init = 0
+    land = 1
+    disarm = 2
+
+class LandTask(AbstractTask):
+
+    def __init__(self, actionvalues_dict):
+        self.tf_buffer = tf2_ros.Buffer()
```
aftersomemath: Fixed

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+        self.tf_listener = tf2_ros.TransformListener(self.tf_buffer)
+        self.canceled = False;
+
+        try:
+            self.LAND_VELOCITY = rospy.get_param('~land_velocity')
+            self.LAND_HEIGHT_TOLERANCE = rospy.get_param('~land_height_tolerance')
+        except KeyError as e:
+            rospy.logerr('Could not lookup a parameter for takeoff task')
```
aftersomemath: Fixed

scripts/land_task.py
aftersomemath: Fixed

scripts/land_task.py
aftersomemath: I agree thats the ideal.

scripts/land_task.py
```diff
@@ -0,0 +1,95 @@
...
+            if self.fc_status != None:
+                # Check that auto pilot is enabled
+                if not self.fc_status.auto_pilot:
+                    return (TaskState.failed, 'flight controller not allowing auto pilot')
+                # Check that the FC is not already armed
+                elif not self.fc_status.armed:
+                    return (TaskState.failed, 'flight controller is not armed, how are we in the air??')
+                # All is good change state to land
+                else:
+                    self.state = LandTaskState.land
```
aftersomemath: The safety coding standards I've used before state to always and an if, else if tree with an else statement to prevent a programmer from creating a situation where no detection is not handled. In our case we happen to return in every other case so it doesn't matter, however I'd rather stick to that convention here.

scripts/land_task.py
aftersomemath: Fixed

scripts/land_task.py
aftersomemath: What?

scripts/land_task.py
aftersomemath: Fixed

scripts/land_task.py
aftersomemath: This is the correct place for a nop. It wasn't because I confused myself earlier.

scripts/motion_planner.py
amiller27: Exceptions from `self.task.cancel`

scripts/motion_planner.py
amiller27: Keep the raise here, just move the signal_shutdown call

scripts/takeoff_task.py
aftersomemath: Fixed

scripts/motion_planner.py
aftersomemath: ah

aftersomemath: Fixed

scripts/motion_planner.py
aftersomemath: Fixed

scripts/land_task.py
amiller27: If the condition three lines above is true, we shouldn't be returning a velocity, we should be going straight out of this block to what we do in the disarm state

scripts/land_task.py
amiller27: Do your safety standards allow for `if`s without `else`s?  Because you actually don't need any `else`s or `elif`s here, and it would be cleaner and more readable without them

scripts/land_task.py
aftersomemath: Fixed

scripts/land_task.py
aftersomemath: Yes they allow if's without else's because a programmer is less likely to end up with a train of if, else ifs, and forget there's a chance it won't be at the bottom. It's a more methodical way of thinking in the end.

amiller27: Then I would ask that you refactor without `elif`s or `else`s to make it more readable

aftersomemath: I believe that is more readable the way it is than without the else. To refactor without the elif would require nested if statements which is less readable.

amiller27: I strongly disagree.  I believe the following is much more readable:
```python
if self._fc_status is None:
    return (TaskState.running, 'velocity', TwistStamped())
if not self._fc_status.auto_pilot:
    return (TaskState.failed, 'flight...')
if not self._fc_status.armed:
    return (TaskState.failed, 'flight...')
self.state = LandTaskState.land
```

amiller27: This makes it glaringly obvious that either the function returns or `self.state` is `LandTaskState.land`

aftersomemath: I see, yes that is better and is a pattern I'm using elsewhere.

aftersomemath: Fixed

scripts/motion_planner.py
amiller27: Why? This isn't a safety thing

aftersomemath: I ended up rolling it into a fix on that branch. I expect to make a PR for the safety client tonight.

amiller27:
Is this ready?  Can you push fixes?

aftersomemath:
LOL it took a little while to fix all my typos after implementing all the changes.

amiller27:
I'm assuming you've tested in the simulator and it works?

aftersomemath:
Yes

PR #33, opened 2017-02-19
Fixed message dependencies in CMakeLists

amiller27:


CMakeLists.txt
```diff
@@ -8,11 +8,10 @@ set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror ${CMAKE_CXX_FLAGS}")
 ## is used, also find other catkin packages
 find_package(catkin REQUIRED COMPONENTS
   actionlib_msgs
-  iarc7_msgs
```
aftersomemath: We use iarc7_msgs::TwistStampedArrayStamped in the acceleration planner. Why is this not needed here?

amiller27: Fixed

PR #38, opened 2017-02-22
Made everything fetch transforms at current time

amiller27:


amiller27:
This has been tested in the simulator and it seems to fix the bug where LowLevelMotion would crash because of transforms timing out.  However, it hasn't been tested on the quad with the props on, so who knows.

amiller27:
Tested on the quad with the props off, it works and fixes (some of) the bugs

param/low_level_motion_0.05.yaml
```diff
@@ -40,3 +40,5 @@ roll_max_rate     : 1.0
 yaw_max           : 3.0
 yaw_min           : -3.0
 yaw_max_rate      : 600.0
+
+update_frequency: 60.0
```
aftersomemath: I thought this was 30hz?

src/LowLevelMotionController.cpp
```diff
@@ -137,8 +131,27 @@ int main(int argc, char **argv)
...
+    // Wait for a valid time in case we are using simulated time (not wall time)
+    while (ros::ok() && ros::Time::now() == ros::Time(0)) {
+        // wait
+        ros::spinOnce();
+    }
+
+    // Create a quad velocity controller. It will output angles corresponding
+    // to our desired velocity
+    double update_frequency;
+    nh.param("update_frequency", update_frequency, 60.0);
```
aftersomemath: I would prefer if this were in the other functions that get parameters to avoid breaking the convention. It's easier on future programmers if the settings are loaded in one spot.

param/low_level_motion_0.05.yaml
amiller27: It was previously limited by the transform frequency.  Now it's not.  robot_localization publishes transforms at 30Hz, as does fc_comms, so I could move this down to 30Hz, but we don't have to

src/LowLevelMotionController.cpp
amiller27: Is it?  The other stuff is a bunch of arrays, so it makes sense to load them in one spot, but for individual variables it just increases the amount of code that needs to be written

param/low_level_motion_0.05.yaml
aftersomemath: I see. There's not a big issue leaving as is. However I'd like to entertain ideas to remove this variable entirely.

amiller27: That doesn't make sense with the current design.  There is nothing limiting the loop rate other than this variable, because the accelerationplanner outputs continuous profiles.

src/LowLevelMotionController.cpp
aftersomemath: Yes it makes it clear where everything is coming from and where it is loaded from. We also set precedent by doing this for the first feed forward parameter

param/low_level_motion_0.05.yaml
aftersomemath: Sorry, you're right. I meant its usage in the quadvelocitycontroller. Not something to work with now though.

amiller27: I agree, it would be nice to eliminate it there

PR #39, opened 2017-03-04
Refactor task commands to objects, allow arbitrary number of task commands.

aftersomemath:


src/iarc7_motion/iarc_tasks/takeoff_task.py
```diff
@@ -9,7 +9,8 @@
 from iarc7_msgs.msg import FlightControllerStatus
 
 from .abstract_task import AbstractTask
-from .task_state import TaskState
+from .task_states import *
```
amiller27: Wildcard imports are bad https://www.python.org/dev/peps/pep-0008/#imports

src/iarc7_motion/motion_planner.py
```diff
@@ -13,7 +13,8 @@
 from iarc7_safety.SafetyClient import SafetyClient
 
 from iarc_task_action_server import IarcTaskActionServer
-from iarc_tasks.task_state import TaskState
+from iarc_tasks.task_states import *
```
amiller27: No wildcard imports

src/iarc7_motion/iarc_tasks/takeoff_task.py
aftersomemath: Fixed

PR #40, opened 2017-03-06
Takeoff land xy

aftersomemath:
Adds position holding to the takeoff and land tasks. Introduces the concept of helper classes that tasks can use to perform particular motions. I'm expecting there will be a good amount of debate about how these helper classes should be implemented.

One thing I noticed in this PR is that the trapezoidal movement profile code could be further abstracted from the PositionHold class. However I feel that that would be appropriate for a second PR when another type of motion is added that requires trapezoidal motion as well.

### DIFF
--- a/scripts/waypoints_test.py
+++ b/scripts/waypoints_test.py
@@ -79,8 +79,6 @@ if __name__ == '__main__':
             velocity.twist.linear.y = constrain((target[1] - trans[1]) * kP, -max_vel, max_vel)
         if abs(target[2] - trans[2]) >= 0.02:
             velocity.twist.linear.z = target[2] - trans[2]
-
-        rospy.logerr('vx %s vy %s x %s y %s', velocity.twist.linear.x, velocity.twist.linear.y, trans[0], trans[1])
         
         # Get the yaw (z axis) rotation from the quanternion
         current_yaw = tf.transformations.euler_from_quaternion(rot, 'rzyx')[0]
--- a/scripts/waypoints_test.py
+++ b/scripts/waypoints_test.py
@@ -79,6 +79,8 @@ if __name__ == '__main__':
             velocity.twist.linear.y = constrain((target[1] - trans[1]) * kP, -max_vel, max_vel)
         if abs(target[2] - trans[2]) >= 0.02:
             velocity.twist.linear.z = target[2] - trans[2]
+
+        rospy.logerr('vx %s vy %s x %s y %s', velocity.twist.linear.x, velocity.twist.linear.y, trans[0], trans[1])
         
         # Get the yaw (z axis) rotation from the quanternion
         current_yaw = tf.transformations.euler_from_quaternion(rot, 'rzyx')[0]
--- a/src/iarc7_motion/iarc_tasks/position_holder.py
+++ b/src/iarc7_motion/iarc_tasks/position_holder.py
@@ -14,8 +14,7 @@ from nav_msgs.msg import Odometry
 class PositionHolder():
     def __init__(self, x=None, y=None):
         update_rate = rospy.get_param('~update_rate', False)
-        #self._update_period = 1.0/update_rate
-        self._update_period = 0.1
+        self._update_period = 1.0/update_rate
         self._lock = threading.RLock()
         self._odometry = None
         self._hold_x = x
@@ -38,6 +37,9 @@ class PositionHolder():
                                                                           z_velocity=z_velocity)
             else:
                 rospy.logerr('get_xy_hold_response called before odometry published')
+                response = TwistStamped()
+                response.header.stamp = rospy.Time.now()
+                response.header.frame_id = 'level_quad'
 
             return response
 
@@ -49,6 +51,8 @@ class PositionHolder():
                     self._hold_x = odometry.pose.pose.position.x
                 if self._hold_y is None:
                     self._hold_y = odometry.pose.pose.position.y
+            else:
+                rospy.logwarn('Received odometry message with incorrect frame and child frames')
 
     def _calculate_trapezoidal_acceleration(self, distance, speed):
         # Check if we can decelerate in time
@@ -62,7 +66,7 @@ class PositionHolder():
             if speed < self._max_speed:
                 target_acceleration = self._max_acceleration
             else:
-                target_acceleration = 0
+                target_acceleration = 0.0
         else:
             target_acceleration = -self._max_acceleration
 
--- a/src/iarc7_motion/iarc_tasks/position_holder.py
+++ b/src/iarc7_motion/iarc_tasks/position_holder.py
@@ -78,20 +78,19 @@ class PositionHolder():
         if speed_towards_target > 0:
             # Check if we can decelerate in time
             try:
-                target_acceleration = (speed_towards_target**2)/(2*distance)
+                required_deceleration = (speed_towards_target**2)/(2*distance)
             except ZeroDivisionError:
                 # We are at the target point, no need to accelerate anywhere
                 return 0.0
 
-            if target_acceleration < self._max_acceleration:
+            if required_deceleration < self._max_acceleration:
                 if self._last_speed < self._max_speed:
-                    target_acceleration = self._max_acceleration
+                    return self._max_acceleration
                 else:
-                    target_acceleration = 0.0
+                    return 0.0
             else:
-                target_acceleration = -self._max_acceleration
+                return -self._max_acceleration
 
-            return target_acceleration
         else:
             # speed is less than zero or zero
             if self._last_speed < self._max_speed:
--- a/src/iarc7_motion/iarc_tasks/position_holder.py
+++ b/src/iarc7_motion/iarc_tasks/position_holder.py
@@ -70,7 +70,7 @@ class PositionHolder():
             # Compute a dot product to find our current speed towards a target
             # The z term can be discarded since we aren't concerned with the z velocity
             speed_towards_target = ((self._last_vel_x * x + self._last_vel_y * y)
-                                    / self._last_speed)
+                                    / distance)
         except ZeroDivisionError:
             # Last speed was zero therefore the last speed to the target was 0
             speed_towards_target = 0
@@ -78,7 +78,7 @@ class PositionHolder():
         if speed_towards_target > 0:
             # Check if we can decelerate in time
             try:
-                target_acceleration = (self._last_speed**2)/(2*distance)
+                target_acceleration = (speed_towards_target**2)/(2*distance)
             except ZeroDivisionError:
                 # We are at the target point, no need to accelerate anywhere
                 return 0.0
@@ -105,7 +105,6 @@ class PositionHolder():
 
         target_acceleration = self._calculate_trapezoidal_acceleration(x, y,
                                                               distance)
-
         current_angle = math.atan2(y, x)
 
         target_twist = TwistStamped()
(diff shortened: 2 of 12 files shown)


src/iarc7_motion/iarc_tasks/position_holder.py
```diff
@@ -0,0 +1,109 @@
...
+
+        return target_acceleration
+
+    def _get_next_acceleration_target_trapezoidal(self, x, y, current_twist, z_velocity=None):
+        # calculate current straight line velocity
+        speed = math.sqrt(current_twist.twist.linear.x**2 + current_twist.twist.linear.y**2)
+        # calculate straight line distance
+        distance = math.sqrt(x**2 + y**2)
+
+        target_acceleration = self._calculate_trapezoidal_acceleration(distance, speed)
```
amiller27: This speed shouldn't actually be the total speed, it should be the projection of our velocity on the vector from our position to (x, y)

src/iarc7_motion/iarc_tasks/position_holder.py
```diff
@@ -0,0 +1,109 @@
...
+            target_acceleration = (speed**2)/(2*distance)
+        except ZeroDivisionError:
+            # We are at the target point, no need to accelerate anywhere
+            return 0.0
+
+        if target_acceleration < self._max_acceleration:
+            if speed < self._max_speed:
+                target_acceleration = self._max_acceleration
+            else:
+                target_acceleration = 0
```
amiller27: This should be 0.0

src/iarc7_motion/iarc_tasks/position_holder.py
```diff
@@ -0,0 +1,109 @@
...
+                                                                          self._odometry.twist,
+                                                                          z_velocity=z_velocity)
+            else:
+                rospy.logerr('get_xy_hold_response called before odometry published')
+
+            return response
+
+    def _current_velocity_callback(self, odometry):
+        with self._lock:
+            if odometry.header.frame_id == 'map' and odometry.child_frame_id == 'level_quad':
```
aftersomemath: Why? Isn't robot localization free to publish different transforms?

amiller27: There shouldn't be odometry for different robots on the same topic (for example each roomba has its odometry published by the simulator on /sim/roomba#/odom).  So in theory yes, we could have multiple robot_localization nodes publishing odometry for different robots, but they should never be on that topic

scripts/waypoints_test.py
```diff
@@ -79,6 +79,8 @@ def constrain(x, l, h):
             velocity.twist.linear.y = constrain((target[1] - trans[1]) * kP, -max_vel, max_vel)
         if abs(target[2] - trans[2]) >= 0.02:
             velocity.twist.linear.z = target[2] - trans[2]
+
+        rospy.logerr('vx %s vy %s x %s y %s', velocity.twist.linear.x, velocity.twist.linear.y, trans[0], trans[1])
```
aftersomemath: This file shouldn't even be here.

src/iarc7_motion/iarc_tasks/position_holder.py
```diff
@@ -0,0 +1,109 @@
...
+import rospy
+
+from geometry_msgs.msg import TwistStamped
+from nav_msgs.msg import Odometry
+
+class PositionHolder():
+    def __init__(self, x=None, y=None):
+        update_rate = rospy.get_param('~update_rate', False)
+        #self._update_period = 1.0/update_rate
+        self._update_period = 0.1
```
aftersomemath: Fixed

src/iarc7_motion/iarc_tasks/position_holder.py
```diff
@@ -0,0 +1,109 @@
...
+        with self._lock:
+            if self._odometry is not None:
+                delta_x = current_x - self._hold_x
+                delta_y = current_y - self._hold_y
+                response = self._get_next_acceleration_target_trapezoidal(delta_x,
+                                                                          delta_y,
+                                                                          self._odometry.twist,
+                                                                          z_velocity=z_velocity)
+            else:
+                rospy.logerr('get_xy_hold_response called before odometry published')
```
aftersomemath: Fixed

src/iarc7_motion/iarc_tasks/position_holder.py
aftersomemath: Could there be odometer for different transforms on the same robot? Or would that also be on a different channel?

src/iarc7_motion/iarc_tasks/position_holder.py
aftersomemath: Fixed

src/iarc7_motion/iarc_tasks/position_holder.py
amiller27: That could happen if you had multiple odometry sensors (maybe multiple imus), and they would go on different topics

src/iarc7_motion/iarc_tasks/position_holder.py
aftersomemath: _calculate_trapezoidal_acceleration is purposefully only concerned with motion in one dimension. The projection is handled below.

src/iarc7_motion/iarc_tasks/position_holder.py
aftersomemath: What if robot localization were concerned with publishing the position of a moving arm that's mounted on top of the quad? In that case the frame_id would be 'quad' and the child_frame 'robot_arm' it's part of the same robot. Is that still a separate topic?

amiller27: Yes, that's a separate topic

aftersomemath: Fixed

aftersomemath:
1. I propose we rename this class to VerticalAscension or something similar. I agree that a position holder should handle x,y, and z discrepancies. However that isn't the purpose of this particular class.

2. I was having problems with that in the simulator already. In that case its not due to noisy signals but poorly tuned PID loops not responding to the velocity request in a timely fashion, but the results are similar. This class should follow a straight line path to the target location. Because of that it would need to be constantly recalculating it's path. I want to wait a bit and see if we can come up with a better solution.

amiller27:
1. I'm suggesting that the purpose of this particular task should be to ramp the z velocity as well, even when taking off it doesn't make sense to require infinite accelerations in the z direction.
2. We talked about this a while back and I thought we concluded that fixed paths were better because of this problem.  Recalculating the path based on current position makes the whole thing susceptible to drift.

src/iarc7_motion/iarc_tasks/position_holder.py
amiller27: I think that we shouldn't be worried about trapezoidal acceleration in the direction perpendicular to our direction of movement, we should request whatever acceleration we need to stay on track in that direction.  This is tied to the larger discussion about path planning in general.

aftersomemath:
The responsibility of the takeoff task is not just to handle the takeoff sequencing but also the fact that takeoff velocity requirements are different than during normal tasks. We can change that but it wouldn't belong in this general class.

It is susceptible to drift right now, but their are also disadvantages to path planning that I don't want to accept.

src/iarc7_motion/iarc_tasks/position_holder.py
aftersomemath: If we were following a path like you proposed below I don't believe this function would be recognizable. For what I tried to do it is the correct implementation.

amiller27:
I'm saying that we should change the takeoff velocity requirements to ramp the velocity.  Yes, they're still different from normal velocity requirements, but I don't think there's ever a situation when we want to have a non-ramped z velocity.

What are the disadvantages to path planning that you're worried about?

aftersomemath:
The ramping of the throttle is specific to the takeoff and landing scenario and should be handled by the task classes specifically. Just like the tasks choose when to use the position hold class or not.

The class is supposed to follow a straight line to it's path not the straight line that existed when it started. I'd like to try and preserve that functionality if possible.

aftersomemath:
What if the position holder used the last velocity it requested instead of the current velocity for creating the next velocity request. Essentially this means a velocity profile will be followed instead of a positional profile.

This means the positional error will always dictate the direction of the next acceleration. However the acceleration will be applied to the last velocity requested instead of the current one. That way if low level motion is not following instructions because of poorly tuned control loops or noise, etc, etc the velocity profile will not be affected.

src/iarc7_motion/iarc_tasks/position_holder.py
aftersomemath: I understand what you meant now. I thought you meant on the output of the function it mattered not the input. (Which also matters and is wrong currently)

aftersomemath:
Velocity targets always point towards the target position. However now the magnitude of the requested vector follows a "profile" of the magnitude of the velocity.

Also the trapezoidal profile uses a projection of the last requested velocity and the error to calculate when to decelerate.

src/iarc7_motion/iarc_tasks/position_holder.py
```diff
@@ -45,62 +47,91 @@ def get_xy_hold_response(self, current_x, current_y, z_velocity=None):
...
+
+        if speed_towards_target > 0:
+            # Check if we can decelerate in time
+            try:
+                target_acceleration = (speed_towards_target**2)/(2*distance)
+            except ZeroDivisionError:
+                # We are at the target point, no need to accelerate anywhere
+                return 0.0
+
+            if target_acceleration < self._max_acceleration:
```
amiller27: Don't we want to leave the target acceleration where it is if it's within limits?

aftersomemath: That is not a limit check. If the target acceleration calculated above is more than the max acceleration it means that we've crossed the point where we can decelerate in time to stop at the target position. I think the code is confusing as written so I'll adjust it.

aftersomemath: Fixed

aftersomemath:
This PR now includes fixing a gimbal lock situation and adjusting the tuning parameters for the sim.

The slew rate and max angle limits are effectively disabled now. It was found that they were greatly affecting the control abilities of the PID controllers. In the case of the throttle the slew rate limit was the only thing stopping the throttle from oscillating wildly from 0-100%. Their effects are disabled now so that the PID loops can be tuned properly.

aftersomemath:
Seeing that the stability issues are corrected. I think this can be merged in once the build is fixed?

amiller27:
Does this last change really give a different number for the yaw than we were getting before?  If you revert just the yaw calculation, it it unstable again?

aftersomemath:
It gives a different yaw calculation under specific circumstances. IIRC prior to finding the transform issue I was able to reproduce the yaw calculation issue using a one axis movement pattern (which was stable due to no yaw being induced). The latest yaw change fixed that issue and caused a stable response.

amiller27:
I still think that we should be ramping the velocity requests in the z direction, but that can wait for another PR if you'd like.

aftersomemath:
Yes that is a different feature IMO

amiller27:
Ok then yes, I think this is ready to merge once it builds (looks like it's just a problem with the dependencies)

amiller27:
Also needs a rebase

PR #41, opened 2017-03-06
Made launch files cooperate with safety information in parameter server

amiller27:
I can rebase this, just wanted to make the PR now

### DIFF
--- a/launch/low_level_motion.launch
+++ b/launch/low_level_motion.launch
@@ -1,9 +1,14 @@
 <launch>
     <arg name="platform" default="0.05" />
+    <arg name="bond_id_namespace" default="safety_bonds" />
+    <arg name="safety_priority" />
 
     <node name="low_level_motion_controller" pkg="iarc7_motion"
         type="low_level_motion_controller">
         <rosparam command="load"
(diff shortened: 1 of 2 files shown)


launch/low_level_motion.launch
```diff
@@ -1,9 +1,14 @@
 <launch>
     <arg name="platform" default="0.05" />
+    <arg name="bond_id_namespace" default="safety_bonds" />
+    <arg name="safety_priority" />
```
aftersomemath: Nodes do not need to know their own safety priority.

launch/low_level_motion.launch
```diff
@@ -1,9 +1,14 @@
...
 
     <node name="low_level_motion_controller" pkg="iarc7_motion"
         type="low_level_motion_controller">
         <rosparam command="load"
             file="$(find iarc7_motion)/param/low_level_motion_$(arg platform).yaml" />
     </node>
+
+    <param name="$(arg bond_id_namespace)/low_level_motion"
```
aftersomemath: Since nodes don't need to know their safety priority this paramater becomes less useful. Instead could we take a parameter for the namespace and a bondId that are married here to create the full bondId that the node needs to use to connect to the monitor?

PR #42, opened 2017-03-08
Filtered velocities

amiller27:


aftersomemath:
Due to our discussions on robots localization output does this result in stability issues with say the way points test? Is their an accompanying PR that adjusts the filtering from localization?

param/low_level_motion_sim.yaml
```diff
@@ -19,26 +19,22 @@ roll_d: 0.0
...
-yaw_accumulator_min: 0.0
-
 throttle_max      : 1.0
 throttle_min      : 0.0
 throttle_max_rate : 1000.0
 
-pitch_max         : 3.5
-pitch_min         : -3.5
-pitch_max_rate    : 1000.0
+pitch_max         : 0.25
```
aftersomemath: I don't think we should turn the max limits or the rate limits back on. They severely affected the stability of the aircraft in the simulator. I think we should have a good discussion about them soon.

param/low_level_motion_sim.yaml
```diff
@@ -19,26 +19,22 @@ roll_d: 0.0
...
+roll_max          : 0.25
+roll_min          : -0.25
+roll_max_rate     : 0.25
 
 yaw_max           : 3.5
 yaw_min           : -3.5
 yaw_max_rate      : 1000.0
 
 update_frequency: 60.0
+max_initial_wait_seconds: 10.0
```
aftersomemath: I think these could use comments explaining what they are for. Perhaps more expressive names too.

src/QuadVelocityController.cpp
```diff
@@ -21,65 +21,81 @@
...
+odometry_subscriber_(nh.subscribe(
+                         "odometry/filtered",
+                         100,
+                         &QuadVelocityController::odometryCallback,
+                         this)),
+odometry_msg_queue_(),
+setpoint_(),
 hover_throttle_(hover_throttle),
-expected_update_frequency_(expected_update_frequency)
+max_initial_wait_(getParam<double>(private_nh, "max_initial_wait_seconds")),
```
aftersomemath: Is it possible to give max_initial_wait_ and max_wait_ more descriptive names?

src/QuadVelocityController.cpp
```diff
@@ -21,65 +21,81 @@
...
+    geometry_msgs::Vector3 velocity;
+    bool success = getVelocityAtTime(velocity, time, max_wait_);
     if (!success) {
         ROS_WARN("Failed to get current velocities in QuadVelocityController::update");
         return false;
     }
 
+    // Get the current transform (rotation) of the quad
+    geometry_msgs::TransformStamped transform;
+    success = getTransformAtTime(transform, time, max_wait_);
```
aftersomemath: Side note: we wouldn't need this and all that goes with it if we used robot localization to publish the orientation.

src/QuadVelocityController.cpp
```diff
@@ -88,48 +104,50 @@ bool QuadVelocityController::update(const ros::Time& time,
...
-    success = throttle_pid_.update(velocity.linear.z, time, throttle_output);
+    success = throttle_pid_.update(velocity.z, time, throttle_output);
     if (!success) {
         ROS_WARN("Throttle PID update failed in QuadVelocityController::update");
         return false;
     }
 
     // Update pitch PID loop
-    success = pitch_pid_.update(velocity.linear.x, time, pitch_output);
+    success = pitch_pid_.update(std::cos(current_yaw) * velocity.x
```
aftersomemath: Would it improve readability to put the transformation code onto the line above with a descriptive variable name saying what frame the vector was transformed to?

src/QuadVelocityController.cpp
```diff
@@ -88,48 +104,50 @@ bool QuadVelocityController::update(const ros::Time& time,
...
+                                time,
+                                pitch_output);
     if (!success) {
         ROS_WARN("Pitch PID update failed in QuadVelocityController::update");
         return false;
     }
 
     // Update roll PID loop
-    success = roll_pid_.update(-velocity.linear.y, time, roll_output);
+    success = roll_pid_.update(std::sin(current_yaw) * velocity.x
```
aftersomemath: Same as variable name for vector transform comment.

src/QuadVelocityController.cpp
```diff
@@ -215,150 +250,103 @@ bool QuadVelocityController::getTransformAtTime(
...
-    double world_x_vel = (transform2.transform.translation.x
-                        - transform1.transform.translation.x) / delta_seconds;
-    double world_y_vel = (transform2.transform.translation.y
-                        - transform1.transform.translation.y) / delta_seconds;
+    throttle_pid_.setSetpoint(setpoint_.linear.z);
 
-    double current_yaw = yawFromQuaternion(transform2.transform.rotation);
+    // Pitch and roll velocities are transformed according to the last yaw
+    // angle because the incoming target velocities are in the map frame
+    pitch_pid_.setSetpoint(setpoint_.linear.x * std::cos(current_yaw)
```
aftersomemath: Transformation put into variable on separate line?

src/QuadVelocityController.cpp
```diff
@@ -215,150 +250,103 @@ bool QuadVelocityController::getTransformAtTime(
...
-
-    result.angular.z = yawChangeBetweenOrientations(
-                            transform1.transform.rotation,
-                            transform2.transform.rotation)
-                     / delta_seconds;
-
-    return result;
+    // Note: Roll is inverted because a positive y velocity means a negative
+    // roll by the right hand rule
+    roll_pid_.setSetpoint(-(setpoint_.linear.x * -std::sin(current_yaw)
```
aftersomemath: Transformation put into variable on separate line?

src/QuadVelocityController.cpp
```diff
@@ -215,150 +250,103 @@ bool QuadVelocityController::getTransformAtTime(
...
+    std::vector<nav_msgs::Odometry>::iterator geq_msg
+        = std::upper_bound(odometry_msg_queue_.begin(),
+                           odometry_msg_queue_.end(),
+                           time,
+                           &QuadVelocityController::timeVsMsgStampedComparator
+                                <nav_msgs::Odometry>);
+
+    // Linear interpolation between next_odom and last_odom
+    const nav_msgs::Odometry& next_odom = *geq_msg;
+    const nav_msgs::Odometry& last_odom = *(geq_msg - 1);
```
aftersomemath: I think it would be good to check geq_msg isn't std:begin before doing this. Even if the logic of the class is supposed to guarantee that this doesn't turn into std::begin what if its accidentally broken in the future?

src/QuadVelocityController.cpp
```diff
@@ -142,36 +160,53 @@ bool QuadVelocityController::update(const ros::Time& time,
...
-    bool success = getTransformAtTime(
-                       transform,
-                       ros::Time(0),
-                       ros::Duration(INITIAL_TRANSFORM_WAIT_SECONDS));
+    const ros::Time start_time = ros::Time::now();
+    while (ros::ok()
+           && odometry_msg_queue_.empty()
+           && ros::Time::now() < start_time + max_initial_wait_) {
+        ros::spinOnce();
+        ros::Duration(0.001).sleep();
```
aftersomemath: On line 212 we use a ros::Duration(0.005).sleep() I don't really care which we use. (I would prefer 0.005 though) but I think it should be consistent.

amiller27:
Yes, there will be an accompanying pr for that.  I went ahead and made this one so it could go through code review because most of it (probably all of it) won't need to change

src/QuadVelocityController.cpp
```diff
@@ -215,150 +250,103 @@ bool QuadVelocityController::getTransformAtTime(
...
-    // Also assumes that we won't have a rotation difference more than pi
-    if (yaw_difference > M_PI)
-    {
-        yaw_difference = yaw_difference - 2 * M_PI;
+    const ros::Time start_time = ros::Time::now();
+    while (ros::ok()
+           && odometry_msg_queue_.back().header.stamp < time
+           && ros::Time::now() < start_time + timeout) {
+        ros::spinOnce();
+        ros::WallDuration(0.001).sleep();
```
aftersomemath: Duration should match other duration sleeps see comment above

src/QuadVelocityController.cpp
```diff
@@ -142,36 +160,53 @@ bool QuadVelocityController::update(const ros::Time& time,
...
 
     if (!success)
     {
         ROS_ERROR("Failed to fetch initial transform");
+        return false;
+    }
+    else
+    {
+        // Mark the last update time as 1ns later than the odometry message,
+        // because we should always have a message older than the last update
```
aftersomemath: I understand why this comment is here. But could we add "so that we will wait for a second orientation"

aftersomemath:
I agree

param/low_level_motion_sim.yaml
amiller27: Any suggestions for more expressive names?  I'm thinking `timeout` and `startup_timeout`?

aftersomemath: odometry_wait_timeout and odometry_wait_startup_timeout?

amiller27: But it's not just a timeout for odometry, it's also for the transforms

aftersomemath: timeout and startup_timeout is fine then

amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Because it's not actually so we wait for a second orientation, it's because the invariant on the odometry buffer is that there's always a message older than the last update time.  I'll put a comment explaining that by the declaration.

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
```diff
@@ -88,48 +104,56 @@ bool QuadVelocityController::update(const ros::Time& time,
...
+    double local_x_velocity = std::cos(current_yaw) * velocity.x
+                            + std::sin(current_yaw) * velocity.y;
+    double local_y_velocity = -std::sin(current_yaw) * velocity.x
+                            +  std::cos(current_yaw) * velocity.y;
+
+    geometry_msgs::TwistStamped velocity_twist;
+    velocity_twist.header.stamp = ros::Time::now();
+    velocity_twist.twist.linear.x = local_x_velocity;
+    velocity_twist.twist.linear.y = local_y_velocity;
+    uav_velocities_.publish(velocity_twist);
```
aftersomemath: Is this supposed to be here?

amiller27: No

PR #44, opened 2017-03-09
Added cmd_vel velocity converter

amiller27:
This takes the movement_velocity_path and converts it into a single velocity published on cmd_vel, which can be used as a control input for robot_localization.  Right now it just picks the first command from the movement_velocity_path and publishes that, could be more sophisticated in the future.


amiller27:
This creates a dependency loop, which may or may not be a problem.  I would want to test this with robot_localization and look at the debug output to make sure there's nothing funky going on.

aftersomemath:
Why wasn't the dependency loop there before. It seems to me that the very fact that robot localization takes the current commanded velocity created a dependency loop.

In this second version the information is coming from the same source except now it's guaranteed to be the actual currently commanded velocity whereas the python script relied on the fact that the first item in the array would be the currently commanded velocity.

amiller27:
You're right, the loop was there before at a higher level in the motion stack.  I tend not to worry about that kind of thing on the higher levels because it's ok if those control loops aren't quite as tight.

I like this better and I'm guessing it'll be ok, I'd just like to test it.  If it's not ok, I'd probably want to keep this version and switch robot_localization to use a Twist instead of a TwistStamped.  Bottom line is that it isn't super important that this velocity is incredibly accurate (because the filter shouldn't rely on this too much), it just needs to give the filter a push in the right direction to get rid of the lag.

aftersomemath:
Ok. I tested with my extreme take off land I was talking about earlier and it worked great. I did not look at any graphs though.

amiller27:
Just tested and as I suspected the control is timing out a lot of the time.  Switching robot_localization to use an unstamped control seems to solve this problem.  It means that robot_localization will be using slightly outdated command velocities (probably around one cycle behind), but that should be ok.

aftersomemath:
So I think it would be good to continue to have the timeout functionality in case things get out of sync, why wouldn't we raise the timeout period?

amiller27:
Because there's a dependency loop, so raising the timeout period doesn't make things any better.  We still have the problem that llm is waiting on a transform from robot_localization which is waiting on a control signal from llm, so raising the timeout will just make robot_localization wait longer.

aftersomemath:
I see. I didn't know robot localization was waiting for a command velocity to be sent.

amiller27:
FYI, it'll actually still time out if it doesn't receive a velocity within the timeout period, it just doesn't have the dependency loop problem

PR #46, opened 2017-03-21
Add thrust model

amiller27:


include/iarc7_motion/QuadVelocityController.hpp
```diff
@@ -22,10 +22,37 @@
 #include "iarc7_msgs/Float64Stamped.h"
 #include "iarc7_msgs/OrientationThrottleStamped.h"
 #include "nav_msgs/Odometry.h"
+#include "std_msgs/Float32.h"
 
 namespace Iarc7Motion
 {
 
+    struct ThrustModel
```
aftersomemath: We should make a trello card to abstract this out.

src/QuadVelocityController.cpp
```diff
@@ -82,13 +88,38 @@ bool QuadVelocityController::update(const ros::Time& time,
...
+    if (!success) {
+        ROS_WARN("Failed to get current transform in QuadVelocityController::update");
+        return false;
+    }
+
+    // Get the current transform (rotation) of the quad
+    geometry_msgs::TransformStamped col_height_transform;
+    success = getTransformAtTime(col_height_transform,
+                                 "quad",
+                                 "level_quad",
```
aftersomemath: Should be a different frame than the one above? This one should be map?

src/QuadVelocityController.cpp
```diff
@@ -189,30 +223,120 @@ bool QuadVelocityController::waitUntilReady()
...
+    battery_msg_queue_.push_back(msg_to_queue);
+
+    // Keep only one message older than the last update time
+    std::vector<iarc7_msgs::Float64Stamped>::const_iterator first_geq_time
+        = std::lower_bound(battery_msg_queue_.begin(),
+                           battery_msg_queue_.end(),
+                           last_update_time_,
+                           &QuadVelocityController::timeVsMsgStampedComparator
+                                <iarc7_msgs::Float64Stamped>);
+    if (first_geq_time - 1 > battery_msg_queue_.begin()) {
```
aftersomemath: Check if this is equal to std::begin before subtracting?

src/QuadVelocityController.cpp
```diff
@@ -189,30 +223,120 @@ bool QuadVelocityController::waitUntilReady()
...
+    }
+}
+
+bool QuadVelocityController::getBatteryAtTime(
+        double& voltage,
+        const ros::Time& time,
+        const ros::Duration& timeout)
+{
+    // Wait until there's a message with stamp >= time
+    if (!waitForBatteryAtTime(time, timeout))
```
aftersomemath: Subtract a delay duration from here to allow less strict timing requirements for battery updates since they are sent much slower than transform updates?

aftersomemath: Pass in the timeout as well as how far back it can look for battery voltages?

include/iarc7_motion/QuadVelocityController.hpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
amiller27: Decided against this

src/QuadVelocityController.cpp
amiller27: Fixed

src/QuadVelocityController.cpp
```diff
@@ -303,10 +309,20 @@ bool QuadVelocityController::getBatteryAtTime(
     // Wait until there's a message with stamp >= time
     if (!waitForBatteryAtTime(time, timeout))
```
aftersomemath: The problem of the battery messages limiting the update rate of the total controller is still here. This function will wait for 0.1 seconds (the update_timeout) parameter. I think this needs to wait for a battery at time (time - battery_timeout).

amiller27: Fixed

aftersomemath:
Let's wait to merge until this is fully tested on the drone.

PR #48, opened 2017-03-26
Fixed bug with TestTask using outdated TaskState stuff

amiller27:


amiller27:
Already tested in the sim, this fixes the bug.  Didn't show up earlier because Python was loading a compiled version of the module.

PR #50, opened 2017-03-29
Add acceleration

amiller27:
Acceleration is now fetched from the `/accel/filtered` topic for the LLM PID controllers instead of estimated by differentiating velocity.  Hasn't been tested yet.


amiller27:
This has been tested, it seems to work

PR #52, opened 2017-04-04
Cleaned up error messages, fixed bad assert

amiller27:
Error messages are now (hopefully) more descriptive.  Also there was a
timing check on a received message that was a little stricter than it
should've been, i.e. it wasn't allowing equal timestamps when it
should've been, which was causing crashes in the simulator when the
timing is coarse.  This check was actually redundant in the first place,
and as such has been replaced with a ROS_ASSERT.


aftersomemath:
Is there supposed to be an update to a timeout somewhere?

PR #55, opened 2017-05-04
Takeoff land procedures rewrite

aftersomemath:
Low level motion now handles both takeoff and land. Because of this we can do some fancy things.

Takeoff is implemented as a controller. This means it has direct access to the "stick" positions. It should not be used to attain high heights because of this. It is for takeoff. Not ascending.

Landing is implemented as a planner. Because of this it can be used for descending and landing. In the future it will be able to be extended to track and land easily. (That' should probably be a separate planner but the pattern is here).

There are some tidbits of things I left in on purpose for the upcoming feature of using the foot switches to detect when we are on the ground.

Todo:
Update the throttle/thrust model so that the recorded hover throttle can be used.
Use foot switches to detect height.

Points of controversy:
* Sending the control input has been moved to the end of a low level motion update because otherwise there would need to be more logic between associating planners with controllers. I don't think this is a big deal but it could be changed if needed.
* During takeoff we send a control input of zero. I'd like a way to remedy this however takeoff is a special case where we don't have a controller. Possible solution is to switch to the velocity controller immediately after the feet leave the ground and not ascend at all.
* We don't have to assert the node when the land planner or the takeoff controller fail during an update because we could fall back to the Quad Velocity controller. However departing from this would mark a fundamental design departure in LLM.


src/iarc7_motion/iarc_tasks/land_task.py
```diff
@@ -16,50 +14,48 @@
                                     TaskFailed)
 from iarc_tasks.task_commands import (VelocityCommand,
```
amiller27: We don't need the velocitycommand anymore

src/iarc7_motion/iarc_tasks/takeoff_task.py
```diff
@@ -106,39 +87,21 @@ def get_desired_command(self):
...
-            if(transStamped.transform.translation.z > self._MIN_MANEUVER_HEIGHT):
-                hold_twist = self._path_holder.get_xy_hold_response(z_velocity=self._TAKEOFF_VELOCITY)
-                return (TaskRunning(), VelocityCommand(hold_twist))
-            else:
-                velocity = TwistStamped()
-                velocity.header.frame_id = 'level_quad'
-                velocity.header.stamp = rospy.Time.now()
-                velocity.twist.linear.z = self._TAKEOFF_VELOCITY
-                return (TaskRunning(), VelocityCommand(velocity))
+            return (TaskRunning(),)
```
amiller27: Do we need a `NopCommand` here?

src/iarc7_motion/motion_planner.py
```diff
@@ -77,6 +90,28 @@ def run(self):
...
+                rospy.logerr(traceback.format_exc())
+                rospy.logwarn('Motion planner aborted task')
+                self._action_server.set_aborted()
+                self._task = None
+            self._ground_interaction_task_callback = None
+        else:
+            rospy.logerr('Ground interaction done callback received with no task callback available')
+
+    def _handle_ground_interaction_command(self, ground_interaction_command):
+        self._ground_interaction_task_callback = ground_interaction_command.completion_callback
```
amiller27: We should check here that we aren't overwriting a pending callback (what if the task requests this again before the first one finishes?)

include/iarc7_motion/TakeoffController.hpp
```diff
@@ -0,0 +1,89 @@
...
+enum class TakeoffState { RAMP,
+                          ASCEND,
+                          DONE };
+
+    class TakeoffController
+    {
+    public:
+        TakeoffController() = delete;
+
+        // Require construction with a node handle and action server
```
aftersomemath: It shouldn't mention an action server. The action server was removed with a later revision.

include/iarc7_motion/TakeoffController.hpp
```diff
@@ -0,0 +1,89 @@
...
+        bool __attribute__((warn_unused_result)) update(
+            const ros::Time& time,
+            iarc7_msgs::OrientationThrottleStamped& uav_command);
+
+        /// Waits until this object is ready to begin normal operation
+        bool __attribute__((warn_unused_result)) waitUntilReady();
+
+        bool isDone();
+
+        double getHoverThrottle();
```
aftersomemath: Sure

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+      last_update_time_(),
+      startup_timeout_(ros_utils::ParamUtils::getParam<double>(
+              private_nh,
+              "startup_timeout")),
+      update_timeout_(ros_utils::ParamUtils::getParam<double>(
+              private_nh,
+              "update_timeout"))
+{
+  // Saving this here to prevent compile warnings. It will be needed in a soon to come revision.
+  ros::NodeHandle nh_ = nh;
```
aftersomemath: Good idea.

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// Used to reset and check initial conditions for takeoff
+// Update needs to begin being called shortly after this is called.
+bool TakeoffController::resetForTakeover(const ros::Time& time)
```
aftersomemath: It's not the same as wait until ready. Sure in this controller they are similar but in quad velocity controller there's a lot of difference.

I'll change this to a better name. Perhaps prepareForTakeover.

src/LandPlanner.cpp
```diff
@@ -0,0 +1,160 @@
...
+                                                    "foot1",
+                                                    "map",
+                                                    time,
+                                                    update_timeout_);
+    if (!success) {
+        ROS_ERROR("Failed to get current transform in LandPlanner::update");
+        return false;
+    }
+
+    if(state_ == LandState::BEGIN_DESCENT) {
```
aftersomemath: Yes

src/TakeoffController.cpp
amiller27: That works

src/LowLevelMotionController.cpp
```diff
@@ -231,36 +265,120 @@ int main(int argc, char **argv)
...
+                    if(success)
+                    {
+                        ROS_DEBUG("Transitioning to takeoff mode");
+                        motion_state = MotionState::TAKEOFF;
+                    }
+                    else
+                    {
+                        ROS_ERROR("Failure transitioning to takeoff mode");
+                        server.setAborted();
+                        motion_state = MotionState::VELOCITY_CONTROL;
```
aftersomemath: We go back to velocity mode if there is a failure to transition into takeoff mode. It's the one case were it seemed safe to allow fault recovery of some kind since the handoff between controller types has not occurred yet.

I like the idea of a GROUNDED mode. I think it would make more sense in another PR though.

src/LowLevelMotionController.cpp
```diff
@@ -231,36 +265,120 @@ int main(int argc, char **argv)
...
+                    quadController.setThrustModel(thrust_model);
+                    ROS_ASSERT_MSG(success, "LowLevelMotion switching to velocity control failed");
+                    motion_state = MotionState::VELOCITY_CONTROL;
+                }
+            }
+            else if(motion_state == MotionState::LAND) {
 
-            // Publish the current target velocity
-            uav_velocity_target_.publish(target_twist);
+                bool success = landPlanner.getTargetTwist(current_time, target_twist);
```
aftersomemath: The land planner would provide the entire twist in the end.

For instance another type of planner might be a land on roomba planner. It needs to take care of XY stuff too.

src/LowLevelMotionController.cpp
amiller27: I think it should be in this PR because the states are a new thing being introduced here in LLM.  I don't think we should start off in velocity mode.

amiller27: I can change it if you don't want to.

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+    if (time < last_update_time_) {
+        ROS_ERROR("Tried to reset TakeoffHandler with time before last update");
+        return false;
+    }
+
+    // Get the current transform (xyz) of the quad
+    geometry_msgs::TransformStamped transform;
+    bool success = transform_wrapper_.getTransformAtTime(transform,
+                                                    "foot1",
```
aftersomemath: It's going away as soon as we have switches.

I'm not sure if this case has been tested, however this is the same order being used for quadvelocity controller. In addition it's working for the other checks.

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+                                                    update_timeout_);
+    if (!success) {
+        ROS_ERROR("Failed to get current transform in TakeoffController::update");
+        return false;
+    }
+
+    if(state_ == TakeoffState::RAMP) {
+        if(transform.transform.translation.z > max_takeoff_start_height_) {
+          recorded_hover_throttle_ = throttle_;
+          throttle_ = std::min(throttle_ + takeoff_throttle_bump_, 100.0);
```
aftersomemath: yes, that bit me somewhere else too but I didn't think to check here.

src/LowLevelMotionController.cpp
amiller27: I agree about the land planner providing the entire twist, but I don't think we need multiple planners, I would rather let the single planner switch frames.

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+          state_ = TakeoffState::ASCEND;
+        }
+        else
+        {
+          throttle_ = std::min(throttle_ + ((time - last_update_time_).toSec() * takeoff_throttle_ramp_rate_), 100.0);
+        }
+    }
+    else if(state_ == TakeoffState::ASCEND)
```
aftersomemath: Fair, this was one of the things I thought might be suggested.

At that point takeoff should be considered complete and control should pass back to HLM to direct LLM and the velocity controller through that procedure.

src/TakeoffController.cpp
amiller27: Yeah I thought I had deleted that second comment before submitting the review, but apparently I hadn't, it's gone now

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+                                                    startup_timeout_);
+    if (!success)
+    {
+        ROS_ERROR("Failed to fetch initial transform");
+        return false;
+    }
+
+    // Mark the last update time the current time as we could not have
+    // received a message before this
+    last_update_time_ = ros::Time::now();
```
aftersomemath: I agree, I'll change 170 to use ros::Time::now() and set last_update_time_ to the same time

src/TakeoffController.cpp
amiller27: Although not all of the places this is being used are going away when we get switches, and they should really all be `base_footprint`

src/iarc7_motion/iarc_tasks/takeoff_task.py
aftersomemath: No you aren't required to return a command if you don't want to. A nop command is a zero velocity command and I don't believe I actually want to send that here.

NopCommand's should probably have a different name.

src/TakeoffController.cpp
amiller27: I disagree, because of the minimum maneuver height.  LLM should be using a velocity controller as soon as we're in the air, but shouldn't allow maneuvers to go all the way down to 0

src/iarc7_motion/motion_planner.py
aftersomemath: yes

src/TakeoffController.cpp
aftersomemath: I'll change to base_footprint

src/iarc7_motion/iarc_tasks/takeoff_task.py
amiller27: Ok, I just noticed that all the other points where `TaskRunning` is returned also tack on a `NopCommand`

src/TakeoffController.cpp
aftersomemath: The minimum maneuver height doesn't apply unless we are maneuvering. It's much cleaner to do this in an HLM task because the infrastructure is there. HLM will call LLM with the takeoff task and then command upwards velocities until we are a fixed amount over the maneuver height.

It also helps keep LLM as a group of low level actions that can be performed instead of high level tasks which contributes to keeping the HLM and LLM less complex.

src/LowLevelMotionController.cpp
aftersomemath: Fair, I'll change it

src/LowLevelMotionController.cpp
aftersomemath: Yeah that might be a better way.

src/TakeoffController.cpp
amiller27: I think I'm ok with that.  The issue is that it makes LLM's interface weird in that clients request takeoff but must then ascend to a certain height before they can request a horizontal velocity.  I know you don't think LLM's interface is important, which I'm assuming means you don't care about that.

aftersomemath: I don't see it that way. LLM let's you do whatever you want really whenever you want. The only times it limits you is if it recognizes what you're doing as physically impossible.

HLM is enforcing the minimum maneuver height with what I proposed. It doesn't allow the user to maneuver if we are below a certain height. It does this by requiring that a special takeoff task be called that takes care of calling the appropriate LLM actions until you are above the appropriate height.

I think the name takeoff becomes misleading because really what LLM is providing is ground interactions. And in the case of takeoff of that interaction is over once we leave the ground.

amiller27: I don't agree that LLM lets (or should let) you do whatever you want whenever you want, HLM is a user of LLM just as much as the abstract is a user of HLM.  However, you're right that if what LLM provides is a "startup" or "initialization" action then it makes sense.

aftersomemath: Fair, in this case I suppose it's true but you're right it might not be true for others.

amiller27: It shouldn't really matter whether it's true for other cases.  Even in this case, it's easier to test and debug if HLM and LLM are two separate units and aren't so interdependent that their functionality and behavior can't be separated.

include/iarc7_motion/QuadVelocityController.hpp
```diff
@@ -52,6 +52,9 @@ namespace Iarc7Motion
         // Set a target velocity for the PID loops
         void setTargetVelocity(geometry_msgs::Twist twist);
 
+        // Use a new thrust model
+        void setThrustModel(ThrustModel thrust_model);
```
aftersomemath: Fixed

include/iarc7_motion/TakeoffController.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "iarc7_msgs/OrientationThrottleStamped.h"
+
+namespace Iarc7Motion
+{
+
+enum class TakeoffState { RAMP,
+                          ASCEND,
+                          DONE };
+
+    class TakeoffController
```
aftersomemath: Fixed

include/iarc7_motion/TakeoffController.hpp
aftersomemath: Fixed

include/iarc7_motion/TakeoffController.hpp
aftersomemath: Fixed

src/TakeoffController.cpp
aftersomemath: Fixed

src/TakeoffController.cpp
aftersomemath: Fixed

src/TakeoffController.cpp
aftersomemath: Fixed

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+    if (time < last_update_time_) {
+        ROS_ERROR("Tried to update TakeoffHandler with time before last update");
+        return false;
+    }
+
+    // Get the current transform (xyz) of the quad
+    geometry_msgs::TransformStamped transform;
+    bool success = transform_wrapper_.getTransformAtTime(transform,
+                                                    "foot1",
```
aftersomemath: Fixed

src/TakeoffController.cpp
aftersomemath: Fixed

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    if(state_ == TakeoffState::RAMP) {
+        if(transform.transform.translation.z > max_takeoff_start_height_) {
+          recorded_hover_throttle_ = throttle_;
+          throttle_ = std::min(throttle_ + takeoff_throttle_bump_, 100.0);
+          state_ = TakeoffState::ASCEND;
+        }
+        else
+        {
+          throttle_ = std::min(throttle_ + ((time - last_update_time_).toSec() * takeoff_throttle_ramp_rate_), 100.0);
```
aftersomemath: Fixed

src/TakeoffController.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    last_update_time_ = time;
+    return true;
+}
+
+bool TakeoffController::waitUntilReady()
+{
+    geometry_msgs::TransformStamped transform;
+    bool success = transform_wrapper_.getTransformAtTime(transform,
+                                                    "foot1",
```
aftersomemath: Fixed

src/TakeoffController.cpp
aftersomemath: Fixed

include/iarc7_motion/LandPlanner.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+// ROS message headers
+#include "geometry_msgs/TwistStamped.h"
+
+#include "iarc7_motion/GroundInteractionAction.h"
+
+namespace Iarc7Motion
+{
+
+typedef actionlib::SimpleActionServer<iarc7_motion::GroundInteractionAction> Server;
```
aftersomemath: Fixed

include/iarc7_motion/LandPlanner.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+enum class LandState { BEGIN_DESCENT,
+                       DESCEND,
+                       BRACE_FOR_IMPACT,
+                       DONE };
+
+    class LandPlanner
```
aftersomemath: Fixed

src/LandPlanner.cpp
```diff
@@ -0,0 +1,160 @@
...
+      startup_timeout_(ros_utils::ParamUtils::getParam<double>(
+              private_nh,
+              "startup_timeout")),
+      update_timeout_(ros_utils::ParamUtils::getParam<double>(
+              private_nh,
+              "update_timeout"))
+{
+  // Saving this here to prevent compile warnings. It will be needed in a soon to come revision.
+  // That involves the landing gear
+  ros::NodeHandle nh_ = nh;
```
aftersomemath: Fixed

src/LandPlanner.cpp
aftersomemath: Fixed

src/LandPlanner.cpp
```diff
@@ -0,0 +1,160 @@
...
+                                                    time,
+                                                    update_timeout_);
+    if (!success) {
+        ROS_ERROR("Failed to get current transform in LandPlanner::update");
+        return false;
+    }
+
+    if(state_ == LandState::BEGIN_DESCENT) {
+        requested_z_vel_ = std::max(requested_z_vel_ - ((time - last_update_time_).toSec() * descend_acceleration_), descend_rate_);
+        if(requested_z_vel_ >= descend_rate_) {
```
aftersomemath: Fixed

src/LandPlanner.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    last_update_time_ = time;
+    return true;
+}
+
+bool LandPlanner::waitUntilReady()
+{
+    geometry_msgs::TransformStamped transform;
+    bool success = transform_wrapper_.getTransformAtTime(transform,
+                                                    "foot1",
```
aftersomemath: Fixed

src/LandPlanner.cpp
```diff
@@ -0,0 +1,160 @@
...
+                                                    startup_timeout_);
+    if (!success)
+    {
+        ROS_ERROR("Failed to fetch initial transform");
+        return false;
+    }
+
+    // Mark the last update time the current time as we could not have
+    // received a message before this
+    last_update_time_ = ros::Time::now();
```
aftersomemath: Fixed

src/iarc7_motion/iarc_tasks/land_task.py
aftersomemath: Fixed

src/iarc7_motion/iarc_tasks/takeoff_task.py
aftersomemath: I actually want nops after those. But not in this case

src/iarc7_motion/motion_planner.py
aftersomemath: Fixed

src/TakeoffController.cpp
aftersomemath: Actually in quad velocity controller you just the get max updated time of the 3 interpolation classes. I'm going to leave this as ros::Time::now and leave it as a comment. The time is only used for the calculation of the ramping speed. I will leave a clarifying comment. Same goes for the LandPlanner.

include/iarc7_motion/LandPlanner.hpp
amiller27: I meant the other stuff shouldn't be indented, that's what I've been doing and I thought that was our convention

include/iarc7_motion/TakeoffController.hpp
amiller27: I meant the other stuff shouldn't be indented, that's what I've been doing and I thought that was our convention

include/iarc7_motion/LandPlanner.hpp
aftersomemath: Well then we weren't following "our" convention in most of the files lol. Fixed again

src/TakeoffController.cpp
aftersomemath: Fixed.

aftersomemath:
All comments are hopefully addressed.

src/iarc7_motion/iarc_tasks/task_commands.py
```diff
@@ -5,7 +5,7 @@
 from geometry_msgs.msg import TwistStamped
 
 class VelocityCommand(object):
-    def __init__(self,  target_twist=None):
```
amiller27: Are we sure nothing else relies on this default arg?

amiller27: We're good, I checked.

aftersomemath: Yes I checked too. I removed it so that it couldn't be used as a fake nop command anymore. You have to construct a twiststamped which would be unproductive.

PR #67, opened 2017-06-07
Track roomba task

asaba96:
Created TrackRoombaTask that takes in a frameID for a roomba and uses a P-controller to have the drone move closer/track the roomba. The task will continue tracking until the drone is within a specified x and y distance to the roomba, which this distance is set in the params let as roomba_max_acceptable_dist. 

Created helper class HeightHolder that determines a hold response to maintain a height at which roomba tracking is to take place. This parameter is also added to the params list. 

Added FrameID to the QuadMove.action so that the requested roomba frame id can be set for track roomba task 

I have tested the current _k_x and _k_y values and seem to track the roomba very well in the sim. In addition, the _k_z in the HeightHolder seems to work well for maintaining the drone's height.


asaba96:
@Riteshlogin Added comments to explain what each method does

amiller27:
Not a complete review, just a comment: could you rebase onto master instead of merging master into your branch?  It makes the history much cleaner (plus it's a good thing to learn, it's how most git workflows operate).  I can show you how tonight if you want.

asaba96:
@amiller27 Yeah my bad

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -0,0 +1,37 @@
...
+            rospy.logerr('Could not lookup a parameter for track roomba task')
+            raise
+
+        while self._odometry is None:
+            pass
+
+    #uses a p-controller to return a velocity to maintain a height
+    #that is set as the param _track_roomba_height
+    def get_height_hold_response(self):
+        delta_z = self._TRACK_HEIGHT - self._odometry.pose.pose.position.z
```
aftersomemath: Because there are two threads modifying the state of variables in this object we should probably use locks at the start of this function and current velocity callback.

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -0,0 +1,37 @@
...
+
+    #uses a p-controller to return a velocity to maintain a height
+    #that is set as the param _track_roomba_height
+    def get_height_hold_response(self):
+        delta_z = self._TRACK_HEIGHT - self._odometry.pose.pose.position.z 
+        response = self._k_z * delta_z + self._odometry.twist.twist.linear.z
+        return response
+
+    def _current_velocity_callback(self, odometry):
+        self._odometry = odometry
```
aftersomemath: Lock here as well. Also need a new line at the end of the file

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+class TrackRoombaTask(AbstractTask):
+
+    def __init__(self, task_dictionary):
+
+        self._roomba_id = task_dictionary['frame_id']
+
+        self._roomba_id = self._roomba_id  + '/base_link'
+        
+        #p-terms
+        self._k_x = .35
```
aftersomemath: Can k_x and k_y be parameters?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+        self._k_y = .35
+
+        if self._roomba_id is None or len(self._roomba_id) < 2:
```
aftersomemath: Why are you checking the length of the roomba id? It could be any length as we have not specified a format for id's. Just that they will be presumably non-empty strings.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+            raise
+
+        if self._TRACK_HEIGHT < self._MIN_MANEUVER_HEIGHT:
+            raise ValueError('Track Roomba height was below the minimum maneuver height')
+
+        self._z_holder = HeightHolder()
+
+        self._state = TrackObjectTaskState.init
+
+        while self._roomba_array is None:
```
aftersomemath: Same comment I made for the height holder. Lets use the convention of translate stop planner for now

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+            pass
+
+    def _receive_roomba_status(self, data):
```
aftersomemath: Also needs a lock

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+        self._roomba_array = data
+
+    def get_desired_command(self):
```
aftersomemath: lock

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+                return (TaskAborted(msg='Exception when looking up transform during roomba track'),)
+
+            #Creat point centered at drone's center
+            stamped_point = PointStamped()
+            stamped_point.point.x = 0
+            stamped_point.point.y = 0
+            stamped_point.point.z = 0
+
+            #returns point distances of roomba to center point of level quad
+            self._roomba_point =tf2_geometry_msgs.do_transform_point(
```
aftersomemath: space after '='

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+                                                stamped_point, roomba_transform)
+
+            if not self._check_max_roomba_range():
+                return (TaskAborted(msg='The provided roomba is not found or not within a meter of the quad'),)
+
+            if self._check_min_roomba_range():
+                return (TaskDone(),)
+
+            #p-controller
```
aftersomemath: Sorry, really nitpicky but could you put spaces between the comment sign and the first letter of the comment. You commented consistently in this style everywhere. I think it makes it more readable. (Up for debate)

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+            return (TaskRunning(), VelocityCommand(velocity)) 
+
+    #checks to see if passed in roomba id is available and
+    #that the drone and roomba are both within a 1 meter square
+    def _check_max_roomba_range(self):
+        for odometry in self._roomba_array.data:
+            if odometry.child_frame_id == self._roomba_id:
+                self._roomba_odometry = odometry
+                self._roomba_found =  True 
+        return (abs(self._roomba_point.point.x) < 1.0 and
```
aftersomemath: Can this be changed to be a radius instead of a square. Can each term be on it's own line for readability?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+                                                stamped_point, roomba_transform)
+
+            if not self._check_max_roomba_range():
+                return (TaskAborted(msg='The provided roomba is not found or not within a meter of the quad'),)
+
+            if self._check_min_roomba_range():
+                return (TaskDone(),)
+
+            #p-controller
+            x_error = (self._roomba_point.point.x * self._k_x +
```
aftersomemath: Can we rename these variables to x_vel_target, y_vel_target, z_vel_target etc? Or something to indicate that they aren't the error term. The error term is the self._roomba_point.point

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+
+            #p-controller
+            x_error = (self._roomba_point.point.x * self._k_x + 
+                        self._roomba_odometry.twist.twist.linear.x)
+            y_error = (self._roomba_point.point.y * self._k_y + 
+                        self._roomba_odometry.twist.twist.linear.y)
+            z_error = self._z_holder.get_height_hold_response()
+
+            #caps x and y velocities
+            if abs(x_error) > self._MAX_TRANSLATION_SPEED:
```
aftersomemath: Can we instead scale the resulting vector to the length of the max translation speed including the z request? Doing it this way means that our maximum velocity is actually much about 70% above _MAX_TRANSLATION_SPEED (though that would require that the z velocity be capped which it isn't so technically we could end up going infinitely faster)

You can do this by calculating the length of the desired translation vector using the Pythagorean method in 3 dimensions to get the magnitude vel_target. If that value is more than the max translation speed you multiply each term like so. x_vel_target = x_vel_target * (max speed / vel_target).

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+            if odometry.child_frame_id == self._roomba_id:
+                self._roomba_odometry = odometry
+                self._roomba_found =  True 
+        return (abs(self._roomba_point.point.x) < 1.0 and 
+                abs(self._roomba_point.point.y) < 1.0 and self._roomba_found)
+
+    #checks the radial distance from the drone to the roomba
+    def _check_min_roomba_range(self):
+        _distance_to_roomba = (self._roomba_point.point.x**2 + 
+                            self._roomba_point.point.y**2)**(.5)
```
aftersomemath: Pretty please math.sqrt for readability?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+
+        while self._roomba_array is None:
+            pass
+
+    def _receive_roomba_status(self, data):
+        self._roomba_array = data
+
+    def get_desired_command(self):
+        if (self._state == TrackObjectTaskState.init or 
+            self._state == TrackObjectTaskState.waiting):
```
aftersomemath: Because the waiting state is here you will switch out of the waiting state and then back into it on line 93. Can you instead use an if statement after line 92 to transition from the waiting state to the track state? It would be cleaner I think.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+            self._state = TrackObjectTaskState.track
+
+        if self._canceled:
+            return (TaskCanceled(),)
+
+        if self._roomba_array is None:
```
aftersomemath: Feeds on previous comment maybe:
if self._state = TrackObjectTaskState.waiting:
    if note have roombas
        do something
    else
        transition state
else
// last if to satisfy the dangling else problem

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -0,0 +1,168 @@
...
+            self._TRANSFORM_TIMEOUT = rospy.get_param('~transform_timeout')
+            self._MIN_MANEUVER_HEIGHT = rospy.get_param('~min_maneuver_height')
+            self._TRACK_HEIGHT = rospy.get_param('~track_roomba_height')
+            self._MAX_TRANSLATION_SPEED = rospy.get_param('~max_translation_speed')
+            self._MAX_ACCEPTABLE_DIST = rospy.get_param('~roomba_max_acceptable_dist')
+        except KeyError as e:
+            rospy.logerr('Could not lookup a parameter for track roomba task')
+            raise
+
+        if self._TRACK_HEIGHT < self._MIN_MANEUVER_HEIGHT:
```
aftersomemath: Can we check this continually within the main function call as well as here?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: @amiller27 and I talked about how the ID's were going to be formatted and they should never be smaller than 2 characters. I added this as a check against invalid id's but it might not be necessary, so I can remove it if desired.

aftersomemath: I would agree that it's invalid because we might in theory want to check things that aren't roombas. Basically we shouldn't check a format that's not finalized. Maybe we should error if we can't find the error in the transform tree?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: done

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: Done. fixed to match other styles

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: done

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: done

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: Done

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: done

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: it errors out already if it cannot lookup the transform so I took the length Check out

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: This whole section had wacky logic because I was always changing it so I refactored it.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: same. I refactored this section

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -0,0 +1,37 @@
...
+
+import math
+import rospy
+
+from geometry_msgs.msg import TwistStamped
+from nav_msgs.msg import Odometry
+
+class HeightHolder():
+    def __init__(self):
+        self._k_z = .1 #p-term for z-height
```
asaba96: done

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -0,0 +1,37 @@
...
+        self._current_velocity_sub = rospy.Subscriber('/odometry/filtered',
+                                              Odometry,
+                                              self._current_velocity_callback)
+        try:
+            self._TRACK_HEIGHT = rospy.get_param('~track_roomba_height')
+        except KeyError as e:
+            rospy.logerr('Could not lookup a parameter for track roomba task')
+            raise
+
+        while self._odometry is None:
```
asaba96: this is partially addressed with the refactored state checking and I removed the while loop. I think with the new state checking logic this should be handled much better, or at least in a more readable form.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: I was thinking that it would be better as well. so fixed

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: I meant to have this height check in height holder, so adding it there. However, what this is doing is making sure we didn't mess up our params and have a track roomba height lower than min maneuver height.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: while loop is gone

asaba96:
@aftersomemath I added the locks, but please check that I did them correctly

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -6,32 +6,40 @@
...
-    #that is set as the param _track_roomba_height
+    # uses a p-controller to return a velocity to maintain a height
+    # that is set as the param _track_roomba_height
     def get_height_hold_response(self):
-        delta_z = self._TRACK_HEIGHT - self._odometry.pose.pose.position.z 
-        response = self._k_z * delta_z + self._odometry.twist.twist.linear.z
-        return response
+        with self._lock:
+            if self._odometry.pose.pose.position.z  < self._MIN_MANEUVER_HEIGHT:
+                raise ValueError('Drone height was below the minimum maneuver height')
```
aftersomemath: Don't need to raise a value error, instead you could return a TaskAborted for instanced.

src/iarc7_motion/iarc_tasks/height_checker.py
```diff
@@ -0,0 +1,37 @@
...
+import math
+import rospy
+import threading
+
+from geometry_msgs.msg import TwistStamped
+from nav_msgs.msg import Odometry
+
+class HeightChecker():
+    def __init__(self):
+        self._lock = threading.RLock()
```
aftersomemath: You don't need locks in here because you aren't using multithreading.

aftersomemath: You don't need locks in here because you aren't using multithreading.

src/iarc7_motion/iarc_tasks/height_checker.py
```diff
@@ -0,0 +1,37 @@
...
+            raise
+            
+    def above_min_maneuver_height(self, current_height):
+        with self._lock:
+            if (current_height < self._MIN_MANEUVER_HEIGHT):
+                return False
+            else:
+                return True
+
+    def check_z_error(self, current_height, desired_height):
```
aftersomemath: Maybe this appropriate for the height holder? Not the height checker.

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -27,19 +25,14 @@ def __init__(self):
...
     # that is set as the param _track_roomba_height
-    def get_height_hold_response(self):
+    def get_height_hold_response(self, height, velocity):
         with self._lock:
-            if self._odometry.pose.pose.position.z  < self._MIN_MANEUVER_HEIGHT:
-                raise ValueError('Drone height was below the minimum maneuver height')
-
-            delta_z = self._TRACK_HEIGHT - self._odometry.pose.pose.position.z 
-            response = self._K_Z * delta_z + self._odometry.twist.twist.linear.z
+            self._current_height = height
```
aftersomemath: Why do you need to save this off?

src/iarc7_motion/iarc_tasks/height_checker.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
```
aftersomemath: Maybe you want to call this the height settings checker

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -57,42 +60,54 @@ def __init__(self, task_dictionary):
...
     def get_desired_command(self):
         with self._lock:
             if (self._state != TrackObjectTaskState.track):
-                if self._roomba_array is None:
+                if self._roomba_array is None or self._drone_odometry is None:
                     self._state = TrackObjectTaskState.waiting
                     return (TaskRunning(), NopCommand())
                 else:
                     self._state = TrackObjectTaskState.track
+            elif not (self._height_checker.above_min_maneuver_height(
```
aftersomemath: Why not use the height checker?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -57,42 +60,54 @@ def __init__(self, task_dictionary):
...
+                        self._drone_odometry.pose.pose.position.z)):
+                return (TaskAborted(msg='Drone is too low'),)
+            elif not (self._height_checker.check_z_error(
```
aftersomemath: Seeing how you use this, I don't think this should be in the height checker. It's kinda specific to this task.

src/iarc7_motion/iarc_tasks/height_checker.py
asaba96: fixed

src/iarc7_motion/iarc_tasks/height_checker.py
asaba96: fixed

asaba96: fixed

src/iarc7_motion/iarc_tasks/height_checker.py
asaba96: I moved it

src/iarc7_motion/iarc_tasks/height_holder.py
asaba96: gone. removed other redundant variables too

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: I did

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: I moved it to height holder.

PR #70, opened 2017-06-20
Hold task

asaba96:
The position hold task uses a p-controller that can either take in an x, y, and z positions to hold (with hold_current_position set to false) or set hold_current_position to true to hold the drone's current position. Since this task is more time dependent, it does not return a TaskDone, but rather depends on your abstract to cancel the task when it is done holding the drone's position. 

There are several checks in place to ensure safe, acceptable inputs and outputs, including velocity limiters. In addition, to avoid abusing this task for translation, it will not run if the distance to the requested point is too large.

### DIFF
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -51,3 +51,7 @@ k_position_x: 1.25
 k_position_y: 1.25
 # p-term for z height for position holder task
 k_position_z: 1.25
+# Hold position is considered finished if x,y,z is within this range of target
+finished_holding_range: .1
+# maximum radial distance from target poimnt at which hold position task is allowed to run
+max_holding_range: 1
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -42,7 +42,7 @@ k_term_tracking_x: .35
 # p-term for Y direction when tracking roombas 
 k_term_tracking_y: .35
 # p-term for height holder helper for Z height
-k_term_hold_z: .90
+p_term_height_hold_z: .90
 # maximum z error for heigth holder
 max_z_error: .7
 # p-term for x height for position holder task
@@ -51,5 +51,5 @@ k_position_x: 1.25
 k_position_y: 1.25
 # p-term for z height for position holder task
 k_position_z: .90
-# maximum radial distance from target poimnt at which hold position task is allowed to run
+# maximum radial distance from target point at which hold position task is allowed to run
 max_holding_range: 1
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -24,7 +24,7 @@ desired_translation_acceleration: 1.0
 # Max translation speed in the xy plane in m/s
 max_translation_speed: 3.0
 #maximum z velocity 
-max_z_velocity: .5
+max_z_velocity: .3
 # Tolerance for a position hold in meters
 translation_position_hold_tolerance: 0.05
 # Translation is considered finished if x,y,z within this range of target
@@ -42,7 +42,7 @@ k_term_tracking_x: .35
 # p-term for Y direction when tracking roombas 
 k_term_tracking_y: .35
 # p-term for height holder helper for Z height
-k_term_hold_z: 1.25
+k_term_hold_z: .90
 # maximum z error for heigth holder
 max_z_error: .7
 # p-term for x height for position holder task
@@ -50,8 +50,6 @@ k_position_x: 1.25
 # p-term for y height for position holder task
 k_position_y: 1.25
 # p-term for z height for position holder task
-k_position_z: 1.25
-# Hold position is considered finished if x,y,z is within this range of target
-finished_holding_range: .1
+k_position_z: .90
 # maximum radial distance from target poimnt at which hold position task is allowed to run
 max_holding_range: 1
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -41,7 +41,13 @@ roomba_max_end_task_dist: .1
 k_term_tracking_x: .35
 # p-term for Y direction when tracking roombas 
 k_term_tracking_y: .35
-# p-term for holding Z height
+# p-term for height holder helper for Z height
 k_term_hold_z: 1.25
-#maximum z error for heigth holder
+# maximum z error for heigth holder
 max_z_error: .7
+# p-term for x height for position holder task
+k_position_x: 1.25
+# p-term for y height for position holder task
+k_position_y: 1.25
+# p-term for z height for position holder task
+k_position_z: 1.25
--- a/src/iarc7_motion/iarc_tasks/hold_position_task.py
+++ b/src/iarc7_motion/iarc_tasks/hold_position_task.py
@@ -81,14 +81,14 @@ class HoldPositionTask(AbstractTask):
                 if self._drone_odometry is None:
                     self._state = HoldPositionTaskStates.waiting
                     return (TaskRunning(), NopCommand())
-                else:
-                    self._set_targets()
-                    return (TaskRunning(), NopCommand())
-
+                
+                self._set_targets()
+                self._state = HoldPositionTaskStates.holding
+                
                 if not self._check_max_error():
                     return (TaskAborted(msg='Desired position is too far away, task not for translation.'),)
 
-                self._state = HoldPositionTaskStates.holding
+                return (TaskRunning(), NopCommand())
 
             elif (self._state == HoldPositionTaskStates.holding):
                 if not (self._height_checker.above_min_maneuver_height(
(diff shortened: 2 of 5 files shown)


param/motion_planner.yaml
```diff
@@ -41,7 +41,15 @@ roomba_max_end_task_dist: .1
 k_term_tracking_x: .35
 # p-term for Y direction when tracking roombas 
 k_term_tracking_y: .35
-# p-term for holding Z height
-k_term_hold_z: 1.25
-#maximum z error for heigth holder
+# p-term for height holder helper for Z height
+k_term_hold_z: .90
```
aftersomemath: Can we name this p_term_height_hold_z

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,160 @@
...
+        self._canceled = False
+
+        self._lock = threading.RLock()
+        self._height_checker = HeightChecker()
+
+        self._current_velocity_sub = rospy.Subscriber(
+            '/odometry/filtered', Odometry,
+            self._current_velocity_callback)
+
+        if not self._hold_current_position:
```
aftersomemath: Can we move this up to around line 41 so we don't have to have this if statement twice?

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,160 @@
...
+                                    TaskCanceled,
+                                    TaskAborted,
+                                    TaskFailed)
+from iarc_tasks.task_commands import (VelocityCommand,
+                                      ArmCommand,
+                                      NopCommand)
+
+from height_checker import HeightChecker
+
+class HoldPostionTaskStates:
```
aftersomemath: Spelling error

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,160 @@
...
+                else:
+                    self._set_targets()
+                    self._state = HoldPostionTaskStates.holding
+            elif not (self._height_checker.above_min_maneuver_height(
+                        self._drone_odometry.pose.pose.position.z)):
+                return (TaskAborted(msg='Drone is too low'),)
+            elif not (self._check_max_error()):
+                return (TaskAborted(msg='Position holder is not for translating'),)
+
+            if self._canceled:
```
aftersomemath: This logic is wrong because we have the possibility of an early return on line 83.

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,160 @@
...
+
+        self._state = HoldPostionTaskStates.init
+
+    def _current_velocity_callback(self, data):
+        with self._lock:
+            self._drone_odometry = data
+
+    def get_desired_command(self):
+        with self._lock:
+            if (self._state != HoldPostionTaskStates.holding):
```
aftersomemath: Can this logic be simplified? For instance have an if statement for each state and do what you need for each state in the if statement. If anything is common to all states it can be done out side of the main state if else block.

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,160 @@
...
+            elif not (self._height_checker.above_min_maneuver_height(
+                        self._drone_odometry.pose.pose.position.z)):
+                return (TaskAborted(msg='Drone is too low'),)
+            elif not (self._check_max_error()):
+                return (TaskAborted(msg='Position holder is not for translating'),)
+
+            if self._canceled:
+                return (TaskCanceled(),)
+
+            if not self._check_max_start_hold_range():
```
aftersomemath: For instance this would be in the if(state==init) block so it's clear when this should run.

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,160 @@
...
+            velocity.twist.linear.y = y_vel_target
+            velocity.twist.linear.z = z_vel_target
+            
+            return (TaskRunning(), VelocityCommand(velocity)) 
+            
+    def cancel(self):
+        rospy.loginfo('HoldPosition Task canceled')
+        self._canceled = True
+
+    def _check_max_error(self):
```
aftersomemath: This function is exactly the same as _check_max_start_hold_range. We don't need two functions to do the same thing.

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,160 @@
...
+    def _check_max_start_hold_range(self):
+        x_vel_target = (self._x_position - self._drone_odometry.pose.pose.position.x)
+        y_vel_target = (self._y_position - self._drone_odometry.pose.pose.position.y)
+        z_vel_target = (self._z_position - self._drone_odometry.pose.pose.position.z)
+
+        _distance_to_point = math.sqrt(x_vel_target**2 + y_vel_target**2 + z_vel_target**2)
+        
+        return (_distance_to_point <= self._MAX_RANGE)
+
+    def _check_inputs(self):
```
aftersomemath: You don't need to do both of the checks on z since min maneuver height should always be positive and parameter checking should be done elsewhere to ensure that the parameter is positive.

Also I don't think you need a function for this because you're only checking one variable.

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,160 @@
...
+        if (self._z_position <= 0 or not 
+            self._height_checker.above_min_maneuver_height(self._z_position)):
+            return False
+        else: 
+            return True
+
+    def _set_targets(self):
```
aftersomemath: You can do this in the odometery topic callback.

src/iarc7_motion/iarc_tasks/hold_position_task.py
asaba96: fixed

src/iarc7_motion/iarc_tasks/hold_position_task.py
asaba96: lolz fixed

param/motion_planner.yaml
asaba96: Ok fixed, but this is the p-term for the z height in the height holder helper. Position holder task and height holder helper use different p-terms for z-height

src/iarc7_motion/iarc_tasks/hold_position_task.py
asaba96: So I moved the check for the max start holding range into the constructor where it belongs. the check max error is for during holding a position, where something is preventing the drone from reaching its target.

src/iarc7_motion/iarc_tasks/hold_position_task.py
asaba96: I simplified this logic, so please review it in my next commit

src/iarc7_motion/iarc_tasks/hold_position_task.py
asaba96: I lied, I fixed this logic

src/iarc7_motion/iarc_tasks/hold_position_task.py
asaba96: oops my bad. Fixed

src/iarc7_motion/iarc_tasks/hold_position_task.py
asaba96: fixed

src/iarc7_motion/iarc_tasks/hold_position_task.py
asaba96: I want to set the hold position as the initial position when the task starts, so thats why I put it separate so I can call it when I am doing my state checking

aftersomemath: Right you can do that in the odometry callback. If self._x_position is none then set it. Otherwise don't.

src/iarc7_motion/iarc_tasks/hold_position_task.py
```diff
@@ -0,0 +1,148 @@
...
+                    self._state = HoldPositionTaskStates.waiting
+                    return (TaskRunning(), NopCommand())
+                else:
+                    self._set_targets()
+
+                if not self._check_max_error():
+                    return (TaskAborted(msg='Desired position is too far away, task not for translation.'),)
+
+                self._state = HoldPositionTaskStates.holding
+
```
aftersomemath: I would prefer if everything below is in an elif block that checks for HoldPositionTaskStates.holding . If we hit the bottom of the function then we should throw an error because an illegal state occurred.

asaba96: fixed

PR #76, opened 2017-07-09
Add the PID debugging, fix parameter loading, tune PID throttle

aftersomemath:


src/QuadVelocityController.cpp
```diff
@@ -192,7 +192,8 @@ bool QuadVelocityController::update(const ros::Time& time,
     success = throttle_pid_.update(velocity.z(),
                                    time,
                                    vertical_accel_output,
-                                   accel.z());
+                                   accel.z(),
+                                   true);
```
amiller27: Can we not merge this as `true`?

PR #75, opened 2017-07-08
Hit roomba task

asaba96:
Created Hit Roomba Task that uses a p-controller to keep the drone centered on the roomba while the drone slowly descends. 
Created Acceleration Limiter helper
Added acceleration limiting to track roomba
Refactored state logic in Track Roomba
Misc. fixes and changes

PR #79, opened 2017-07-13
Velocity test

asaba96:
Created Velocity Test task that returns an x and y velocity that is set in motion_planner.yaml. It also uses a p-controller to keep the drone at its current height.

### DIFF
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -57,7 +57,9 @@ max_holding_range: 1.0
 hit_roomba_max_start_dist: .5
 # descent velocity when hitting a roomba 
 hit_descent_velocity: -0.2
-# test task velocity in x, m/s
-test_velocity_x: 1.0
-# test task velocity in y, m/s
-test_velocity_y: 1.0
+# x velocity (m/s) for velocity task
+velocity_x: 1.0
+# y velocity (m/s) for velocity task
+velocity_y: 1.0
+# height to hold drone at when in velocity task
+velocity_task_height: 1.5
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -58,8 +58,8 @@ hit_roomba_max_start_dist: .5
 # descent velocity when hitting a roomba 
 hit_descent_velocity: -0.2
 # x velocity (m/s) for velocity task
-velocity_x: 1.0
+velocity_x: .5
 # y velocity (m/s) for velocity task
-velocity_y: 1.0
+velocity_y: .5
 # height to hold drone at when in velocity task
 velocity_task_height: 1.5
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -57,3 +57,5 @@ max_holding_range: 1.0
 hit_roomba_max_start_dist: .5
 # descent velocity when hitting a roomba 
 hit_descent_velocity: -0.2
+# test task velocity, m/s
+test_velocity: 1.0
--- a/param/motion_planner.yaml
+++ b/param/motion_planner.yaml
@@ -57,5 +57,7 @@ max_holding_range: 1.0
 hit_roomba_max_start_dist: .5
 # descent velocity when hitting a roomba 
 hit_descent_velocity: -0.2
-# test task velocity, m/s
-test_velocity: 1.0
+# test task velocity in x, m/s
+test_velocity_x: 1.0
+# test task velocity in y, m/s
+test_velocity_y: 1.0
--- a/src/iarc7_motion/iarc_tasks/height_holder.py
+++ b/src/iarc7_motion/iarc_tasks/height_holder.py
@@ -15,7 +15,7 @@ class HeightHolder(object):
     def __init__(self, desired_height):
         self._lock = threading.RLock()
         self._delta_z = 0
-        self._DESIRED_HEIGHT
+        self._DESIRED_HEIGHT = desired_height
         try:
             self._MIN_MANEUVER_HEIGHT = rospy.get_param('~min_maneuver_height')
             self._MAX_Z_ERROR = rospy.get_param('~max_z_error')
--- a/src/iarc7_motion/iarc_tasks/height_holder.py
+++ b/src/iarc7_motion/iarc_tasks/height_holder.py
@@ -11,39 +11,27 @@ import threading
 from geometry_msgs.msg import TwistStamped
 from nav_msgs.msg import Odometry
 
-class HeightHolder():
-    def __init__(self, use_current = False):
+class HeightHolder(object):
+    def __init__(self, desired_height):
         self._lock = threading.RLock()
         self._delta_z = 0
-        self._USE_CURRENT_HEIGHT = use_current
-        self._current_height = None
+        self._DESIRED_HEIGHT
         try:
             self._MIN_MANEUVER_HEIGHT = rospy.get_param('~min_maneuver_height')
             self._MAX_Z_ERROR = rospy.get_param('~max_z_error')
-            self._TRACK_HEIGHT = rospy.get_param('~track_roomba_height')
             self._K_Z = rospy.get_param('~p_term_height_hold_z')
         except KeyError as e:
             rospy.logerr('Could not lookup a parameter for track roomba task')
             raise
 
-        if self._TRACK_HEIGHT < self._MIN_MANEUVER_HEIGHT:
-            raise ValueError('Track Roomba Param height was below the minimum maneuver height')
+        if self._DESIRED_HEIGHT < self._MIN_MANEUVER_HEIGHT:
+            raise ValueError('Desired height was below the minimum maneuver height')
 
     # uses a p-controller to return a velocity to maintain a height
-    # that is set as the param _track_roomba_height
     def get_height_hold_response(self, height):
         with self._lock:
-            if self._current_height is None:
-                self._current_height = height
-
-            if not self._USE_CURRENT_HEIGHT:
-                self._delta_z = self._TRACK_HEIGHT - height
-            else:
-                self._delta_z = self._current_height - height
-
-            response = self._K_Z * self._delta_z
-            
-            return response
+            self._delta_z = self._DESIRED_HEIGHT - height
+            return self._K_Z * self._delta_z
 
     def check_z_error(self, current_height):
-        return (self._delta_z < self._MAX_Z_ERROR)
+        return (abs(self._delta_z) < self._MAX_Z_ERROR)
--- a/src/iarc7_motion/iarc_tasks/height_holder.py
+++ b/src/iarc7_motion/iarc_tasks/height_holder.py
@@ -12,8 +12,11 @@ from geometry_msgs.msg import TwistStamped
 from nav_msgs.msg import Odometry
 
 class HeightHolder():
-    def __init__(self):
+    def __init__(self, use_current = False):
         self._lock = threading.RLock()
+        self._delta_z = 0
+        self._USE_CURRENT_HEIGHT = use_current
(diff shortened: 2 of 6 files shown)


src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -30,13 +33,17 @@ def __init__(self):
     # that is set as the param _track_roomba_height
     def get_height_hold_response(self, height):
         with self._lock:
-            delta_z = self._TRACK_HEIGHT - height
-            response = self._K_Z * delta_z
+            if self._current_height is None:
+                self._current_height = height
```
amiller27: Can we just set the track height here?  It'll make this stuff more readable

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -30,13 +33,17 @@ def __init__(self):
...
+            
             return response
 
     def check_z_error(self, current_height):
-        z_error = abs(current_height-self._TRACK_HEIGHT)
-        if (z_error > self._MAX_Z_ERROR):
-            return False
-        else:
-            return True
+        return (self._delta_z < self._MAX_Z_ERROR)
```
amiller27: We want to check the `abs` of this, right?

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -12,8 +12,11 @@
 from nav_msgs.msg import Odometry
 
 class HeightHolder():
-    def __init__(self):
+    def __init__(self, use_current = False):
```
amiller27: While we're here, could we make this inherit from `object`?

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+                                    TaskFailed)
+from iarc_tasks.task_commands import (VelocityCommand,
+                                      ArmCommand,
+                                      NopCommand)
+
+from height_holder import HeightHolder
+from height_settings_checker import HeightSettingsChecker
+from acceleration_limiter import AccerlationLimiter
+
+class VelocityTestTaskState:
```
amiller27: Inherit from `object`?

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+
+                x_vel_target = self._TRANSLATION_SPEED_X
+                y_vel_target = self._TRANSLATION_SPEED_Y
+                z_vel_target = self._z_holder.get_height_hold_response(
+                    self._drone_odometry.pose.pose.position.z)
+
+                if (abs(z_vel_target) > self._MAX_Z_VELOCITY):
+                    z_vel_target = z_vel_target/abs(z_vel_target) * self._MAX_Z_VELOCITY
+
+                desired_vel = []
```
amiller27: Put these in the list above

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+                desired_vel.append(x_vel_target)
+                desired_vel.append(y_vel_target)
+                desired_vel.append(z_vel_target)
+
+                curr_vel = []
```
amiller27: Same as above

param/motion_planner.yaml
```diff
@@ -52,4 +52,22 @@ k_position_y: 1.25
 # p-term for z height for position holder task
 k_position_z: .90
 # maximum radial distance from target point at which hold position task is allowed to run
-max_holding_range: 1
+max_holding_range: 1.0
```
aftersomemath: Looks like the changes to this file belong with block or hit roomba?

src/iarc7_motion/iarc_tasks/height_holder.py
```diff
@@ -30,13 +33,17 @@ def __init__(self):
     # that is set as the param _track_roomba_height
     def get_height_hold_response(self, height):
         with self._lock:
-            delta_z = self._TRACK_HEIGHT - height
-            response = self._K_Z * delta_z
+            if self._current_height is None:
+                self._current_height = height
+
+            if not self._USE_CURRENT_HEIGHT:
+                self._delta_z = self._TRACK_HEIGHT - height
```
aftersomemath: We don't need current height or track height. We only need something called hold_height. When the class is created the height to hold can be passed into the constructor. The param that loads the roomba track height doesn't really belong here IMO.

We could compromise and allow the track height param to be loaded in the constructor. If a height is passed to the constructor then the new variable hold_height can be set to the passed in height or it can be set to the track height parameter.

aftersomemath: Additionally, the way this works now, get_height_hold_response does two different things without really indicating whats going on.

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+            self._TRANSLATION_SPEED_X = rospy.get_param('~test_velocity_x')
+            self._TRANSLATION_SPEED_Y = rospy.get_param('~test_velocity_y')
+
+        except KeyError as e:
+            rospy.logerr('Could not lookup a parameter for constant velocity test task')
+            raise
+
+        self._z_holder = HeightHolder(True)
+        self._height_checker = HeightSettingsChecker()
+        self._limiter = AccerlationLimiter()
```
aftersomemath: Spelling? Also this can't be merged until #75 is because this is used.

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+                            self._drone_odometry.pose.pose.position.z)):
+                    return (TaskAborted(msg='Drone is too low'),)
+
+                x_vel_target = self._TRANSLATION_SPEED_X
+                y_vel_target = self._TRANSLATION_SPEED_Y
+                z_vel_target = self._z_holder.get_height_hold_response(
+                    self._drone_odometry.pose.pose.position.z)
+
+                if (abs(z_vel_target) > self._MAX_Z_VELOCITY):
+                    z_vel_target = z_vel_target/abs(z_vel_target) * self._MAX_Z_VELOCITY
```
aftersomemath: Consider using copysign here for readability? https://stackoverflow.com/questions/1986152/why-doesnt-python-have-a-sign-function/1986776#1986776

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+
+                desired_vel = []
+                desired_vel.append(x_vel_target)
+                desired_vel.append(y_vel_target)
+                desired_vel.append(z_vel_target)
+
+                curr_vel = []
+                curr_vel.append(self._drone_odometry.twist.twist.linear.x)
```
aftersomemath: Current velocity should only be set to the drones actual velocity on the first run. After that we should use the last requested velocity as the current velocity. Otherwise if the velocity controller is lagging the drone could take a long time to get started.

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self._current_velocity_sub = rospy.Subscriber(
+            '/odometry/filtered', Odometry,
+            self._current_velocity_callback)
+
+        try:
+            self._TRANSFORM_TIMEOUT = rospy.get_param('~transform_timeout')
+            self._MAX_TRANSLATION_SPEED = rospy.get_param('~max_translation_speed')
+            self._MAX_Z_VELOCITY = rospy.get_param('~max_z_velocity')
+            self._TRANSLATION_SPEED_X = rospy.get_param('~test_velocity_x')
```
aftersomemath: I don't think these should be parameters. These should be passed in through the action request.

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+from height_holder import HeightHolder
+from height_settings_checker import HeightSettingsChecker
+from acceleration_limiter import AccerlationLimiter
+
+class VelocityTestTaskState:
+    init = 0
+    moving = 1
+    waiting = 2
+
+class VelocityTestTask(AbstractTask):
```
aftersomemath: Can we rename this to VelocityTask? It's not just for testing

src/iarc7_motion/iarc_tasks/velocity_test_task.py
```diff
@@ -0,0 +1,128 @@
...
+                velocity.header.stamp = rospy.Time.now()
+                velocity.twist.linear.x = desired_vel[0]
+                velocity.twist.linear.y = desired_vel[1]
+                velocity.twist.linear.z = desired_vel[2]
+                
+                return (TaskRunning(), VelocityCommand(velocity)) 
+            else:
+                return (TaskAborted(msg='Illegal state reached in Velocity test task'),)
+
+    def cancel(self):
```
aftersomemath: The only way for this task to exit is by it being canceled. I was hoping that the action message that starts this could set a time period to maintain the velocity for.

param/motion_planner.yaml
asaba96: neither. I wanted to make it a double when I noticed it while writing this task

aftersomemath: You added tons of extra parameters below this line.

asaba96: oh that refers to just line 55. The rest I do not know how to fix

src/iarc7_motion/iarc_tasks/height_holder.py
asaba96: fixed

asaba96: fixed

param/motion_planner.yaml
aftersomemath: In this case it might be easiest to just make a fix commit to delete the extra parameters before it gets merged. Aaron might know a better way though.

amiller27: The better way would be to run git blame to figure out when these were changed and do an interactive rebase to remove them

src/iarc7_motion/iarc_tasks/height_holder.py
asaba96: fixed

src/iarc7_motion/iarc_tasks/height_holder.py
asaba96: I changed it to where you pass in a desired height. Its much more readable

src/iarc7_motion/iarc_tasks/height_holder.py
asaba96: Fixed with a minor rewrite to allow a passed in desired height

asaba96: This is fixed with a minor rewrite of height holder

src/iarc7_motion/iarc_tasks/velocity_task.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python
```
aftersomemath: I was hoping that a time could be set for this task to maintain the velocity for then exit.

aftersomemath: But its fine for now. There are some more important things to do

PR #87, opened 2017-07-24
Height recovery task and beginning of Roomba Rewrite

asaba96:
Height Recovery task that recovers drone's height above min man height. Also began rewrite of Roomba Tracking by adding mode to track roomba

### DIFF
--- a/action/QuadMove.action
+++ b/action/QuadMove.action
@@ -3,6 +3,7 @@ string movement_type
 string frame_id
 bool preempt
 bool hold_current_position
+bool mode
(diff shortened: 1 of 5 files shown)


action/QuadMove.action
```diff
@@ -3,6 +3,7 @@ string movement_type
 string frame_id
 bool preempt
 bool hold_current_position
+bool mode
```
aftersomemath: What does mode mean?

asaba96: fixed

PR #91, opened 2017-07-25
Check for order of messages

Riteshlogin:
Go through the velocity commands and make sure that they came in order

### DIFF
--- a/src/AccelerationPlanner.cpp
+++ b/src/AccelerationPlanner.cpp
@@ -105,7 +105,7 @@ bool AccelerationPlanner::trimVelocityQueue(TwistStampedArray& twists, const ros
     // keep all going into the future.
 
     // Find the first time more than or equal
-    // Uses a lambda to compare the current time and a headers time stamp
+    // Uses a lambda (and an auto specifier) to compare the current time and a header's time stamp
     TwistStampedArray::iterator it = std::lower_bound(
                     twists.begin(),
                     twists.end(),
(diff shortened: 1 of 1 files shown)


src/AccelerationPlanner.cpp
```diff
@@ -181,10 +181,23 @@ bool AccelerationPlanner::appendVelocityQueue(TwistStampedArray& current_twists,
 }
 
 // Receive a new list of velocities commands and call appendVelocityQueue to insert them into the queue
-// TODO: check that message is sorted properly, default to safe state if anythings wrong?
+// The messages in the array should arrive in chronological order (from message[0] to message[n])
 void AccelerationPlanner::processVelocityCommand(const iarc7_msgs::TwistStampedArray::ConstPtr& message)
 {
 
+    TwistStampedArray::const_iterator checkMessageOrder;
```
aftersomemath: Can you check that all the line lengths below are 80 characters or less?

src/LowLevelMotionController.cpp
```diff
@@ -251,6 +251,8 @@ int main(int argc, char **argv)
 
     ros::Rate rate (update_frequency);
 
+    iarc7_msgs::OrientationThrottleStamped last_uav_command;
```
aftersomemath: It looks like you rebased and managed to include a commit that shouldn't be here.

src/AccelerationPlanner.cpp
```diff
@@ -181,14 +181,29 @@ bool AccelerationPlanner::appendVelocityQueue(TwistStampedArray& current_twists,
...
+
+    for(checkMessageOrder = message->twists.begin(); checkMessageOrder != message->twists.end() - 1;
```
aftersomemath: I think a more readable style is to put each of these statements on their own line. Instead of 2 on the top and one on the bottom.

src/AccelerationPlanner.cpp
```diff
@@ -181,14 +181,29 @@ bool AccelerationPlanner::appendVelocityQueue(TwistStampedArray& current_twists,
 }
 
 // Receive a new list of velocities commands and call appendVelocityQueue to insert them into the queue
-// TODO: check that message is sorted properly, default to safe state if anythings wrong?
+// The messages in the array should arrive in chronological order (from message[0] to message[n])
 void AccelerationPlanner::processVelocityCommand(const iarc7_msgs::TwistStampedArray::ConstPtr& message)
 {
 
+    TwistStampedArray::const_iterator checkMessageOrder;
```
aftersomemath: Does this need to be declared outside of the for loop?

Riteshlogin: I originally thought it wouldn't compile if I tried that, but everything went fine

PR #96, opened 2017-09-24
Changed track roomba

Qisgreat:
I changed the track roomba so that it cancels after a certain amount of time. Also I changed the over shoot to take in a value from the abstract

### DIFF
--- a/action/QuadMove.action
+++ b/action/QuadMove.action
@@ -3,10 +3,10 @@ string movement_type
 string frame_id
 bool preempt
 bool hold_current_position
-bool tracking_mode
 float64 takeoff_height
 float64 time_to_track
-float64 overshoot
+float64 x_overshoot
+float64 y_overshoot
 float64 x_position
 float64 y_position
 float64 z_position
--- a/action/QuadMove.action
+++ b/action/QuadMove.action
@@ -5,6 +5,8 @@ bool preempt
 bool hold_current_position
 bool tracking_mode
 float64 takeoff_height
+float64 time_to_track
+float64 overshoot
 float64 x_position
 float64 y_position
 float64 z_position
--- a/src/iarc7_motion/iarc_tasks/track_roomba_task.py
+++ b/src/iarc7_motion/iarc_tasks/track_roomba_task.py
@@ -71,6 +71,8 @@ class TrackRoombaTask(object, AbstractTask):
             self._TRANSFORM_TIMEOUT = rospy.get_param('~transform_timeout')
             self._MAX_HORIZ_SPEED = rospy.get_param('~max_translation_speed')
             self._MAX_START_TASK_DIST = rospy.get_param('~roomba_max_start_task_dist')
+            if not self._check_max_roomba_range():
+                raise ValueError('The provided roomba is not close enough')
             self._MAX_END_TASK_DIST = rospy.get_param('~roomba_max_end_task_dist')
             self._MAX_Z_VELOCITY = rospy.get_param('~max_z_velocity')
             self._K_X = rospy.get_param('~k_term_tracking_x')
@@ -101,7 +103,7 @@ class TrackRoombaTask(object, AbstractTask):
             
             if self._time_to_track != 0 and (rospy.Time.now() 
                 - self._start_time >= rospy.Duration(self._time_to_track)):
-                # ROS_INFO("Task cancelled due to time out")
+                rospy.loginfo('Timed out')
                 return (TaskDone(),)
 
             if self._canceled:
@@ -163,8 +165,11 @@ class TrackRoombaTask(object, AbstractTask):
                 roomba_velocity = math.sqrt(roomba_x_velocity**2 + roomba_y_velocity**2)
 
                 # The overshoot is taking in the x velocity normalizing it and applying overshoot
-                x_overshoot = roomba_x_velocity/roomba_velocity * self._x_overshoot
-                y_overshoot = roomba_y_velocity/roomba_velocity * self._y_overshoot
+                overshoot = math.sqrt(self._x_overshoot**2 + self._y_roomba**2)
+                x_overshoot = overshoot * math.cos(math.atan(roomba_y_velocity / roomba_x_velocity)
+                                                        + math.atan(self._y_overshoot / self._x_overshoot))
+                y_overshoot = overshoot * math.sin(math.atan(roomba_y_velocity / roomba_x_overshoot )
+                                                        + math.atan(self._y_overshoot / self._x_overshoot))
 
                 # p-controller
                 x_vel_target = ((self._roomba_point.point.x + x_overshoot)
@@ -220,10 +225,12 @@ class TrackRoombaTask(object, AbstractTask):
     # that the drone and roomba are both within a specified distance
     # in order to start/continue the task
     def _check_max_roomba_range(self):
+
         _distance_to_roomba = math.sqrt(self._roomba_point.point.x**2 + 
                             self._roomba_point.point.y**2)
         
-        return (_distance_to_roomba <= self._MAX_START_TASK_DIST + ((self._x_overshoot)**2 + (self._y_overshoot)**2)**(0.5))
+        return (_distance_to_roomba <= self._MAX_START_TASK_DIST)
+
 
     def cancel(self):
         rospy.loginfo('TrackRoomba Task canceled')
--- a/src/iarc7_motion/iarc_tasks/track_roomba_task.py
+++ b/src/iarc7_motion/iarc_tasks/track_roomba_task.py
@@ -84,7 +84,7 @@ class TrackRoombaTask(object, AbstractTask):
         self._height_checker = HeightSettingsChecker()
         self._limiter = AccelerationLimiter()
 
-        if _MAX_TASK_DIST < math.sqrt(self._x_overshoot**2
+        if self._MAX_TASK_DIST < math.sqrt(self._x_overshoot**2
                             + self._y_overshoot**2):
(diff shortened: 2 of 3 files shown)


src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -99,6 +96,12 @@ def _current_velocity_callback(self, data):
 
     def get_desired_command(self):
         with self._lock:
+            if self._start_time is None:
+                self._start_time = rospy.Time.now()
+            
+            if not self._time_to_track == 0 and (rospy.Time.now() - self._start_time >= rospy.Duration(self._time_to_track)):
```
asaba96: Can you make this less than 80 characters long? So make it 2 lines.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -153,13 +156,11 @@ def get_desired_command(self):
...
                     return (TaskAborted(msg='The provided roomba is not found or not close enough to the quad'),)
 
-                if self._check_max_ending_roomba_range():
-                    return (TaskDone(),)
-
                 roomba_x_velocity = self._roomba_odometry.twist.twist.linear.x
                 roomba_y_velocity = self._roomba_odometry.twist.twist.linear.y
                 roomba_velocity = math.sqrt(roomba_x_velocity**2 + roomba_y_velocity**2)
 
+                # The overshoot is back in town
```
asaba96: Update this comment to explain what is going on (i.e. using the direction the roomba is moving)

action/QuadMove.action
```diff
@@ -5,6 +5,8 @@ bool preempt
 bool hold_current_position
 bool tracking_mode
```
aftersomemath: This should be removed maybe? You aren't using it in the abstract anymore.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -99,6 +96,13 @@ def _current_velocity_callback(self, data):
 
     def get_desired_command(self):
         with self._lock:
+            if self._start_time is None:
+                self._start_time = rospy.Time.now()
+            
+            if not self._time_to_track == 0 and (rospy.Time.now() 
+                - self._start_time >= rospy.Duration(self._time_to_track)):
+                return (TaskDone(),)
```
aftersomemath: Can we return with a message that it timed out? Maybe a ros info?

action/QuadMove.action
```diff
@@ -5,6 +5,8 @@ bool preempt
...
 float64 takeoff_height
+float64 time_to_track
+float64 overshoot
```
amiller27: Didn't we want this to be a 2d offset (x and y directions)?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -99,6 +96,13 @@ def _current_velocity_callback(self, data):
 
     def get_desired_command(self):
         with self._lock:
+            if self._start_time is None:
+                self._start_time = rospy.Time.now()
+            
+            if not self._time_to_track == 0 and (rospy.Time.now()
```
amiller27: Use `!=` instead of `not self... == 0`

action/QuadMove.action
aftersomemath: I don't believe so because its in polar form.

amiller27: No I mean in the roomba frame, did we want sideways and forward offsets?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -162,8 +164,8 @@ def get_desired_command(self):
                 roomba_velocity = math.sqrt(roomba_x_velocity**2 + roomba_y_velocity**2)
 
                 # The overshoot is taking in the x velocity normalizing it and applying overshoot
-                x_overshoot = roomba_x_velocity/roomba_velocity * self._overshoot
-                y_overshoot = roomba_y_velocity/roomba_velocity * self._overshoot
+                x_overshoot = roomba_x_velocity/roomba_velocity * self._x_overshoot
```
aftersomemath: I don't think this will work because the unit vector being multiplied by is in the global frame while the x and y vectors are supposed to be in the roomba frame. If you think about x_overshoot and y_overshoot as a function of the roombas heading this gives you an ellipsoidal offset in the global frame. Meaning that if _x_overshoot != _y_overshoot you will get different overshoot amounts relative to the roomba depending on the roombas heading in the global frame.

I think you want to rotate the x and y overshoot vectors from the roomba frame to the global frame and then apply them to the positions in line 171 and 173.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -222,7 +224,7 @@ def _check_max_roomba_range(self):
         _distance_to_roomba = math.sqrt(self._roomba_point.point.x**2 + 
                             self._roomba_point.point.y**2)
         
-        return (_distance_to_roomba <= self._MAX_START_TASK_DIST + self._overshoot)
+        return (_distance_to_roomba <= self._MAX_START_TASK_DIST + ((self._x_overshoot)**2 + (self._y_overshoot)**2)**(0.5))
```
aftersomemath: I think this is over 80 characters

amiller27: Also I don't think this is the condition we want here.  Most likely this should stay like it was before, as a max distance from the center of the roomba regardless of the specified offset.  If this is intended to be a cap on the initial distance from the target point, it should be a max distance to the center of the roomba plus the offset, which is a little different from what it's doing now

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -163,8 +165,11 @@ def get_desired_command(self):
                 roomba_velocity = math.sqrt(roomba_x_velocity**2 + roomba_y_velocity**2)
 
                 # The overshoot is taking in the x velocity normalizing it and applying overshoot
-                x_overshoot = roomba_x_velocity/roomba_velocity * self._x_overshoot
-                y_overshoot = roomba_y_velocity/roomba_velocity * self._y_overshoot
+                overshoot = math.sqrt(self._x_overshoot**2 + self._y_roomba**2)
+                x_overshoot = overshoot * math.cos(math.atan(roomba_y_velocity / roomba_x_velocity)
```
amiller27: You probably need to use `atan2` here instead of `atan`

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -69,13 +71,13 @@ def __init__(self, task_request):
             self._TRANSFORM_TIMEOUT = rospy.get_param('~transform_timeout')
             self._MAX_HORIZ_SPEED = rospy.get_param('~max_translation_speed')
             self._MAX_START_TASK_DIST = rospy.get_param('~roomba_max_start_task_dist')
+            if not self._check_max_roomba_range():
```
aftersomemath: When I was talking about the valueerror I was saying we need to check that the parameters roomba_max_start_task_dist and the magnitude of x_overshoot and y_overshoot are congruent.

I don't think this function can be called yet because self._roomba_point is None

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -218,16 +224,7 @@ def _check_roomba_in_sight(self):
...
-        return (_distance_to_roomba <= self._MAX_START_TASK_DIST + self._overshoot)
-
-    # checks the radial distance from the drone to the roomba
-    # in order to end the task as successful
-    def _check_max_ending_roomba_range(self):
-        _distance_to_roomba = math.sqrt(self._roomba_point.point.x**2 + 
-                            self._roomba_point.point.y**2)
-        
-        return (_distance_to_roomba <= (self._MAX_END_TASK_DIST + self._overshoot))
+        return (_distance_to_roomba <= self._MAX_START_TASK_DIST)
```
aftersomemath: Can we name it _MAX_TASK_DIST since it doesn't apply just at the beginning?

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -37,8 +37,9 @@ class TrackRoombaTask(object, AbstractTask):
     def __init__(self, task_request):
 
         self._roomba_id = task_request.frame_id  + '/base_link'
-        # true means hit roomba, false means block
-        self._tracking_mode = task_request.tracking_mode
+        self._time_to_track = task_request.time_to_track
+        self._x_overshoot = task_request.x_overshoot
+        self._y_overshoot = task_request.y_overshoot
```
aftersomemath: Can we check that these overshoots are possible given the limitations imposed by _MAX_START_TASK_DIST. The check would need to be below line 83. Throw a valueerror exception if it occurs.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
aftersomemath: Since the meaning of _MAX_START_TASK_DIST has changed we need to change the value in the .param file to be what it was plus the value of the original overshoot parameter.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
Qisgreat: Fixed

PR #97, opened 2017-09-29
Add go_to_roomba_task and param for it

amjltc295:
Add go_to_roomba_task
New param: ending_radius_to_roomba 
for how close to get to the roomba in go_to_roomba_task

### DIFF
--- a/action/QuadMove.action
+++ b/action/QuadMove.action
@@ -10,6 +10,7 @@ float64 y_position
 float64 z_position
 float64 x_velocity
 float64 y_velocity
+float64 ending_radius_to_roomba # How close to get to the roomba for go_to_roomba
 ---
 #result definition
 bool success
--- a/src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
+++ b/src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python
+
+import math
+import rospy
+import tf2_ros
+import tf2_geometry_msgs
+import threading
+
+from geometry_msgs.msg import TwistStamped
+from geometry_msgs.msg import PointStamped
+from geometry_msgs.msg import Point
+from iarc7_msgs.msg import OdometryArray
+from nav_msgs.msg import Odometry
+from geometry_msgs.msg import Vector3Stamped
+from geometry_msgs.msg import Vector3
+
+from .abstract_task import AbstractTask
+from iarc_tasks.task_states import (TaskRunning,
+                                    TaskDone,
+                                    TaskCanceled,
+                                    TaskAborted,
+                                    TaskFailed)
+from iarc_tasks.task_commands import (VelocityCommand,
+                                      ArmCommand,
+                                      NopCommand)
+
+from height_holder import HeightHolder
+from height_settings_checker import HeightSettingsChecker
+from acceleration_limiter import AccelerationLimiter
+
+class GoToObjectTaskState(object):
+    init = 0
+    going = 1
+    waiting = 2
+
+class GoToRoombaTask(object, AbstractTask):
+
+    def __init__(self, task_request):
+
+        self._roomba_id = task_request.frame_id  + '/base_link'
+        # true means hit roomba, false means block
+        self._ending_radius_to_roomba = task_request.ending_radius_to_roomba
+
+        if self._roomba_id is None:
+            raise ValueError('A null roomba id was provided')
+
+        self._roomba_odometry = None
+        self._roomba_array = None
+        self._roomba_point = None
+        self._roomba_found = False
+        self._current_velocity = None
+
+        self._drone_odometry = None
+        self._canceled = False
+
+        self._tf_buffer = tf2_ros.Buffer()
+        self._tf_listener = tf2_ros.TransformListener(self._tf_buffer)
+
+        self._lock = threading.RLock()
+
+        self._roomba_status_sub = rospy.Subscriber(
+            'roombas', OdometryArray,
+            self._receive_roomba_status)
+
+        self._current_velocity_sub = rospy.Subscriber(
+            '/odometry/filtered', Odometry,
+            self._current_velocity_callback)
+
+        try:
+            self._TRANSFORM_TIMEOUT = rospy.get_param('~transform_timeout')
+            self._MAX_HORIZ_SPEED = rospy.get_param('~max_translation_speed')
+            self._MAX_START_TASK_DIST = rospy.get_param('~roomba_max_start_task_dist')
+            self._MAX_END_TASK_DIST = rospy.get_param('~roomba_max_end_task_dist')
+            self._MAX_Z_VELOCITY = rospy.get_param('~max_z_velocity')
+            self._K_X = rospy.get_param('~k_term_tracking_x')
+            self._K_Y = rospy.get_param('~k_term_tracking_y')
+            GO_TO_HEIGHT = rospy.get_param('~track_roomba_height')
+            _roomba_diameter = rospy.get_param('~roomba_diameter')
+            _drone_width = rospy.get_param('~drone_landing_gear_width')
+        except KeyError as e:
+            rospy.logerr('Could not lookup a parameter for go_to roomba task')
+            raise
+
+        self._z_holder = HeightHolder(GO_TO_HEIGHT)
+        self._height_checker = HeightSettingsChecker()
+        self._limiter = AccelerationLimiter()
+
+        self._overshoot = 0.0
+
+        self._state = GoToObjectTaskState.init
+
+    def _receive_roomba_status(self, data):
+        with self._lock:
+            self._roomba_array = data
+
+    def _current_velocity_callback(self, data):
+        with self._lock:
+            self._drone_odometry = data
+
+    def get_desired_command(self):
+        with self._lock:
+            if self._canceled:
+                return (TaskCanceled(),)
+
+            elif (self._state == GoToObjectTaskState.init):
+                if self._roomba_array is None or self._drone_odometry is None:
+                    self._state = GoToObjectTaskState.waiting
+                else:
+                    self._state = GoToObjectTaskState.going
+                return (TaskRunning(), NopCommand())
+
+            elif (self._state == GoToObjectTaskState.waiting):
+                if self._roomba_array is None or self._drone_odometry is None:
+                    self._state = GoToObjectTaskState.waiting
+                else:
+                    self._state = GoToObjectTaskState.going
+                return (TaskRunning(), NopCommand())
+
+            elif self._state == GoToObjectTaskState.going:
+
+                if not (self._height_checker.above_min_maneuver_height(
+                            self._drone_odometry.pose.pose.position.z)):
+                    return (TaskAborted(msg='Drone is too low'),)
+                elif not (self._z_holder.check_z_error(
+                    self._drone_odometry.pose.pose.position.z)):
+                    return (TaskAborted(msg='Z error is too high'),)
+                elif not self._check_roomba_in_sight():
+                    return (TaskAborted(msg='The provided roomba is not in sight of quad'),)
+
+                try:
+                    roomba_transform = self._tf_buffer.lookup_transform(
+                                        'level_quad',
+                                        self._roomba_id,
+                                        rospy.Time.now(),
+                                        rospy.Duration(self._TRANSFORM_TIMEOUT))
+                except (tf2_ros.LookupException,
+                        tf2_ros.ConnectivityException,
+                        tf2_ros.ExtrapolationException) as ex:
+                    rospy.logerr('ObjectGoToTask: Exception when looking up transform')
+                    rospy.logerr(ex.message)
+                    return (TaskAborted(msg='Exception when looking up transform during go_to roomba'),)
+
+                # Creat point centered at drone's center
+                stamped_point = PointStamped()
+                stamped_point.point.x = 0
+                stamped_point.point.y = 0
+                stamped_point.point.z = 0
+
+                # returns point distances of roomba to center point of level quad
+                self._roomba_point = tf2_geometry_msgs.do_transform_point(
+                                                    stamped_point, roomba_transform)
+
+                """ This is for tracking
+                if not self._check_max_roomba_range():
+                    return (TaskAborted(msg='The provided roomba is not found or not close enough to the quad'),)
+                """
+
+                if self._check_max_ending_roomba_range():
+                    return (TaskDone(),)
+
+                roomba_x_velocity = self._roomba_odometry.twist.twist.linear.x
+                roomba_y_velocity = self._roomba_odometry.twist.twist.linear.y
+                roomba_velocity = math.sqrt(roomba_x_velocity**2 + roomba_y_velocity**2)
+
+                x_overshoot = roomba_x_velocity/roomba_velocity * self._overshoot
+                y_overshoot = roomba_y_velocity/roomba_velocity * self._overshoot
+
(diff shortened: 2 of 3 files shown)


action/QuadMove.action
```diff
@@ -10,6 +10,7 @@ float64 y_position
 float64 z_position
 float64 x_velocity
 float64 y_velocity
+float64 ending_radius_to_roomba # How close to get to the roomba for go_to_roomba
```
asaba96: Not sure what effects the comment makes here for the move actions, but you don't need it.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+                                    TaskFailed)
+from iarc_tasks.task_commands import (VelocityCommand,
+                                      ArmCommand,
+                                      NopCommand)
+
+from height_holder import HeightHolder
+from height_settings_checker import HeightSettingsChecker
+from acceleration_limiter import AccelerationLimiter
+
+class GoToObjectTaskState(object):
```
asaba96: can you call this `GoToRoombaTaskState`?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+        try:
+            self._TRANSFORM_TIMEOUT = rospy.get_param('~transform_timeout')
+            self._MAX_HORIZ_SPEED = rospy.get_param('~max_translation_speed')
+            self._MAX_START_TASK_DIST = rospy.get_param('~roomba_max_start_task_dist')
+            self._MAX_END_TASK_DIST = rospy.get_param('~roomba_max_end_task_dist')
+            self._MAX_Z_VELOCITY = rospy.get_param('~max_z_velocity')
+            self._K_X = rospy.get_param('~k_term_tracking_x')
+            self._K_Y = rospy.get_param('~k_term_tracking_y')
+            GO_TO_HEIGHT = rospy.get_param('~track_roomba_height')
+            _roomba_diameter = rospy.get_param('~roomba_diameter')
```
asaba96: You don't need these variables, for any overshoot you may employ, which I am thinking we do not have any overshoot at all, will not be dependent on roomba nor drone diameters.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+                if not (self._height_checker.above_min_maneuver_height(
+                            self._drone_odometry.pose.pose.position.z)):
+                    return (TaskAborted(msg='Drone is too low'),)
+                elif not (self._z_holder.check_z_error(
+                    self._drone_odometry.pose.pose.position.z)):
+                    return (TaskAborted(msg='Z error is too high'),)
+                elif not self._check_roomba_in_sight():
+                    return (TaskAborted(msg='The provided roomba is not in sight of quad'),)
+
+                try:
```
asaba96: @amiller27 the roomba odometry will give us the roomba's map position, no? If so, `self._roomba_array` should have that position. 

So I'm thinking is we use the roomba's global positions, but avoid using any estimates on the roomba's velocity since we may be too far to know which direction its moving. So here lets not find the vector from us to the roomba, but just use the position of the roomba in the map frame.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+                stamped_point = PointStamped()
+                stamped_point.point.x = 0
+                stamped_point.point.y = 0
+                stamped_point.point.z = 0
+
+                # returns point distances of roomba to center point of level quad
+                self._roomba_point = tf2_geometry_msgs.do_transform_point(
+                                                    stamped_point, roomba_transform)
+
+                """ This is for tracking
```
asaba96: Delete these lines if you are not using them

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+                if not self._check_max_roomba_range():
+                    return (TaskAborted(msg='The provided roomba is not found or not close enough to the quad'),)
+                """
+
+                if self._check_max_ending_roomba_range():
+                    return (TaskDone(),)
+
+                roomba_x_velocity = self._roomba_odometry.twist.twist.linear.x
```
asaba96: Since the abstract will be passing in an ending radial distance away from the Roomba, lets not have any overshoot. So delete lines 161-166

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+                roomba_y_velocity = self._roomba_odometry.twist.twist.linear.y
+                roomba_velocity = math.sqrt(roomba_x_velocity**2 + roomba_y_velocity**2)
+
+                x_overshoot = roomba_x_velocity/roomba_velocity * self._overshoot
+                y_overshoot = roomba_y_velocity/roomba_velocity * self._overshoot
+
+                # p-controller
+                x_vel_target = ((self._roomba_point.point.x + x_overshoot)
```
asaba96: The idea for this task is that it would use the `TranslateStopPlanner` class in order to go across the arena. So what I envision is you would create a `TranslateStopPlanner` every iteration of `get_desired_command` and pass in the current position of the roomba in the map frame, possibly with some overshoot that is dependent on the direction the roomba is moving and how far away from it we are, if that is known (which it won't always be known, so maybe we keep that out).

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+                                    * self._K_X + roomba_x_velocity)
+                y_vel_target = ((self._roomba_point.point.y + y_overshoot)
+                                    * self._K_Y + roomba_y_velocity)
+
+                z_vel_target = self._z_holder.get_height_hold_response(
```
asaba96: If using the `TranslateStopPlanner` just pass in `GO_TO_HEIGHT` as the desired z component and it should take care of holding that height. Am I wrong about this @aftersomemath  or @amiller27 ?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+                desired_vel = [x_vel_target, y_vel_target, z_vel_target]
+
+                drone_vel_x = self._drone_odometry.twist.twist.linear.x
+                drone_vel_y = self._drone_odometry.twist.twist.linear.y
+                drone_vel_z = self._drone_odometry.twist.twist.linear.z
+
+                if self._current_velocity is None:
+                    self._current_velocity = [drone_vel_x, drone_vel_y, drone_vel_z]
+
+                desired_vel = self._limiter.limit_acceleration(self._current_velocity, desired_vel)
```
asaba96: if using the `TranslateStopPlanner` you wouldn't need to limit the acceleration, it should do that for you.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+
+    ## checks to see if passed in roomba id is in sight of quad
+    def _check_roomba_in_sight(self):
+        for odometry in self._roomba_array.data:
+            if odometry.child_frame_id == self._roomba_id:
+                self._roomba_odometry = odometry
+                return True
+        return False
+
+    # that the drone and roomba are both within a specified distance
```
asaba96: You don't need this function since you will be very far from the roomba when this task starts.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,238 @@
...
+
+        return (_distance_to_roomba <= self._MAX_START_TASK_DIST + self._overshoot)
+
+    # checks the radial distance from the drone to the roomba
+    # in order to end the task as successful
+    def _check_max_ending_roomba_range(self):
+        _distance_to_roomba = math.sqrt(self._roomba_point.point.x**2 +
+                            self._roomba_point.point.y**2)
+
+        return (_distance_to_roomba <= (self._ending_radius_to_roomba + self._overshoot))
```
asaba96: You named the ending distance variable differently above. Make sure they are the same otherwise you are comparing your distance to who knows what.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
asaba96: The function to delete is `check_max_roomba_range()`

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
amiller27: Yes, the roomba odometry array is currently set up to include positions in the `map` frame, and I think that's probably how it's going to stay.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
amiller27: @aftersomemath would know better.  One thing I will say is that I don't think `GO_TO_HEIGHT` should be the variable name, first of all locals shouldn't be all caps, and second I don't think it's a descriptive name.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
aftersomemath: Do you really wanted to be creating and destroying objects at 10Hz like that if you don't have to?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
aftersomemath: The translate stop planner will hold the desired z height or change it you request a height other than what you currently are at.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
asaba96: @amiller27 and I talked about it before and we didn't see an issue with it.

aftersomemath: The translate stop planner will need to be refactored to support this. It loads a bunch of parameters from the param server and has a subscriber (and thus also deal with locks). It wasn't designed to be created and destroyed rapidly.

asaba96: Or would it be easier to refactor the `TranslateStopPlanner` to support new set points?

asaba96: And replanning

aftersomemath: IMO all the parameter loading and the subscriber need to be taken out if you want to create and destroy objects of the TranslateStopPlanner rapidly.

If that is done the translate stop planner devolves into essentially a method that you have to call a constructor on before it runs. All its parameters must be loaded in the class that uses it and they have to be passed in each time you create the object.

I agree that its more beautiful the way you've proposed. But in my judgement this crosses the line of practical vs nice code.

@amiller27 What do you think?

amiller27: I think the parameter loading is fine, but the way the whole `TranslateStopPlanner` works is probably an issue because a lot of it doesn't make sense if you only have one message on that topic that the subscriber is listening to

aftersomemath: I'm sorry I didn't understand Andrews first comment. Yes it would be easier to refactor the TranslateStopPlanner so that you can give it new setpoints. Creating and destroying it rapidly doesn't work with its current architecture and won't without significant functionality changes.

I do not understand why parameter loading and reloading over and over again is considered ok its incredibly wasteful.

What do you mean be not making sense if it only has one message? @amiller27 I think you're pointing out another reason why it wasn't designed to be created and destroyed rapidly.

I you guys want something that the translate stop planner is not which is a basically a planner math library.

asaba96: Aaron meant that it listens to the drone's current odometry, so for 1/10th of a second it will have old data, until the next one is created with an updated odometry from the task. So I think a refactor to accept new points is in order.

aftersomemath: How can it not operate on old data if its only ran at 10Hz?

aftersomemath: It listens the drone's odometry to determine where it is so it can plan accelerations accordingly. I think you want to rip that part out and just abstract out the trapezoidal planner part.

amiller27: Lots of things are wasteful.  The `TranslateStopPlanner` et. al. are limited by the rate of data coming into them, not the CPU.  I'm refining the meaning of "wasn't designed to be created and destroyed rapidly" by pointing out that the primary reason that's the case is because it doesn't give full functionality from the first message.

asaba96: I'm saying it WILL operate on old data if you just pass in the odometry every time, which isn't good at all.

aftersomemath: I'm claiming the parameter loading has the potential to be severely update rate limiting wasteful. We don't know yet so I believe tests should be done before jumping into this as a software engineering design pattern. It would be very nice if parameters can be loaded willy nilly all the time.

I see what you mean by requiring two odometery messages to enable full functionality. The reason that exists is because that planners algorithim depends on tracking state between each velocity request. Specifically it needs the last velocity request it made. If you use the odometry message each time you might never go anywhere because the controllers aren't perfect.  If you want to change that you need a new type of planner or you need to pass it in like its a math library.

Or you can just support changing the target position, we will need to do testing to make sure the algorithm handles that well.

action/QuadMove.action
amjltc295: Fixed. Where is the place for explaining these variables?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
amjltc295: Fixed

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
amjltc295: These are from Track Roomba Task and used in several places. There is another variable called self._overshoot. I thought _K_X and the rest are for PID control?

asaba96: So the idea of this task was to use the `TranslateStopPlanner` and not use any weird form of PID control that we use elsewhere, as it is not suitable for long distance translations.

action/QuadMove.action
asaba96: Actually, you can move the comment to above the variable. so on line 13

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
amjltc295: Fixed

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
amjltc295: Fixed

PR #98, opened 2017-10-06
Refactoring takeoff and landing

MiscM:
I'm ready for feedback

### DIFF
--- a/src/LandPlanner.cpp
+++ b/src/LandPlanner.cpp
@@ -46,14 +46,12 @@ LandPlanner::LandPlanner(
       update_timeout_(ros_utils::ParamUtils::getParam<double>(
               private_nh,
               "update_timeout")),
-      uav_arm_client_()
+      uav_arm_client_(nh.serviceClient<iarc7_msgs::Arm>("uav_arm"))
 {
     landing_gear_subscriber_ = nh.subscribe("landing_gear_contact_switches",
                                  100,
                                  &LandPlanner::processLandingGearMessage,
                                  this);
-    //Setting up the service client connection for disarm request
-    uav_arm_client_ = nh.serviceClient<iarc7_msgs::Arm>("uav_arm");
 }
 
 // Used to prepare and check initial conditions for landing
@@ -125,21 +123,21 @@ bool LandPlanner::getTargetTwist(const ros::Time& time,
         }
     }
     else if (state_ == LandState::DISARM) {
-      //Sending disarm request to fc_comms
-      iarc7_msgs::Arm srv;
-      srv.request.data = false;
-      //Check if request was succesful
-      if(uav_arm_client_.call(srv)) {
-        if(srv.response.success == false) {
-          ROS_ERROR("Service could not disarm the controller");
-          return false;
+        // Sending disarm request to fc_comms
+        iarc7_msgs::Arm srv;
+        srv.request.data = false;
+        // Check if request was succesful
+        if(uav_arm_client_.call(srv)) {
+            if(srv.response.success == false) {
+                ROS_ERROR("Service could not disarm the controller");
+                return false;
+            }
         }
-      }
-      else {
-        ROS_ERROR("disarming service failed");
-        return false;
-      }
-      state_ = LandState::DONE;
+        else {
+            ROS_ERROR("disarming service failed");
+            return false;
+        }
+        state_ = LandState::DONE;
     }
     else if(state_ == LandState::DONE)
     {
--- a/src/LandPlanner.cpp
+++ b/src/LandPlanner.cpp
@@ -119,25 +119,22 @@ bool LandPlanner::getTargetTwist(const ros::Time& time,
             return false;
         }
         else if(anyPressed(landing_gear_message_)) {
-            state_ = LandState::DISARM;
-        }
-    }
-    else if (state_ == LandState::DISARM) {
-        // Sending disarm request to fc_comms
-        iarc7_msgs::Arm srv;
-        srv.request.data = false;
-        // Check if request was succesful
-        if(uav_arm_client_.call(srv)) {
-            if(srv.response.success == false) {
-                ROS_ERROR("Service could not disarm the controller");
+            // Sending disarm request to fc_comms
+            iarc7_msgs::Arm srv;
+            srv.request.data = false;
+            // Check if request was succesful
+            if(uav_arm_client_.call(srv)) {
+                if(srv.response.success == false) {
+                    ROS_ERROR("Service could not disarm the controller");
+                    return false;
+                }
+            }
+            else {
+                ROS_ERROR("disarming service failed");
                 return false;
             }
+            state_ = LandState::DONE;
         }
-        else {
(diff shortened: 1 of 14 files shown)


src/LandPlanner.cpp
```diff
@@ -118,9 +121,26 @@ bool LandPlanner::getTargetTwist(const ros::Time& time,
             return false;
         }
         else if(anyPressed(landing_gear_message_)) {
-            state_ = LandState::DONE;
```
asaba96: why do we need a separate state for disarm? can't we just go straight into disarming the drone here once we know we are on the ground?

src/LandPlanner.cpp
```diff
@@ -118,9 +121,26 @@ bool LandPlanner::getTargetTwist(const ros::Time& time,
...
+            state_ = LandState::DISARM;
         }
     }
+    else if (state_ == LandState::DISARM) {
```
asaba96: This block is using 2 space tabs. our style is 4 spaces. Please make your indents 4 spaces.

src/TakeoffController.cpp
```diff
@@ -52,12 +52,15 @@ TakeoffController::TakeoffController(
...
                   private_nh,
-                  "takeoff_max_height_switch_pressed"))
+                  "takeoff_max_height_switch_pressed")),
+      uav_arm_client_()
 {
     landing_gear_subscriber_ = nh.subscribe("landing_gear_contact_switches",
                                    100,
                                    &TakeoffController::processLandingGearMessage,
                                    this);
+    //Setting up the service client connection for arm request
```
asaba96: Please put a space after your // in your comments

src/TakeoffController.cpp
```diff
@@ -144,7 +147,24 @@ bool TakeoffController::update(const ros::Time& time,
         return false;
     }
 
-    if(state_ == TakeoffState::RAMP) {
+    if(state_ == TakeoffState::ARM) {
```
asaba96: Same indent issue as above here.

src/iarc7_motion/iarc_tasks/land_task.py
```diff
@@ -30,66 +29,46 @@ def __init__(self, task_request):
...
-            # Check if disarm succeeded
-            if self._disarm_request_success:
-                return (TaskFailed(), NopCommand())
-            else:
-                rospy.logerr('Low level motion action failed, disarming then exiting')
-                return (TaskRunning(), ArmCommand(False, True, False, self.disarm_callback))
+        # Change state to done
+        elif self._state == LandTaskState.done:
+            return (TaskDone(), NopCommand())
+
```
asaba96: delete this extra space

src/iarc7_motion/iarc_tasks/height_recovery_task.py
```diff
@@ -15,7 +15,6 @@
                                     TaskAborted,
```
asaba96: This should be tested in the sim before merging because there are somewhat bigger changes here. running the `hit_roomba` abstract will test the height recovery task.

src/iarc7_motion/iarc_tasks/takeoff_task.py
```diff
@@ -15,21 +15,17 @@
...
 class TakeoffTaskState:
-    init = 0
-    request_arm = 1
-    pause_before_takeoff = 2
-    takeoff = 3
-    ascend = 4
-    ascend_angle = 5
-    done = 6
-    failed = 7
+    init= 0
```
amiller27: Missing a space here

src/iarc7_motion/iarc_tasks/takeoff_task.py
```diff
@@ -129,26 +98,23 @@ def get_desired_command(self):
             if(transStamped.transform.translation.z > self._TAKEOFF_COMPLETE_HEIGHT):
                 self._state = TakeoffTaskState.done
                 return (TaskDone(), NopCommand())
-            elif (self._state == TakeoffTaskState.ascend
-              and transStamped.transform.translation.z
-                > self._ANGLE_MODE_HEIGHT):
-                self._state = TakeoffTaskState.ascend_angle
-                return (TaskRunning(), ArmCommand(True, True, True, lambda _ : None))
+
```
amiller27: Do we not need to verify that the drone is in angle mode?  Is this done in `fc_comms` now?

MiscM: I don't think we have the check, which is probably the reason why we accidentally took off in rate mode during testing on Thursday. I think that should definitely be implemented into fc_comms

src/LandPlanner.cpp
```diff
@@ -45,12 +45,15 @@ LandPlanner::LandPlanner(
...
-              "update_timeout"))
+              "update_timeout")),
+      uav_arm_client_()
 {
     landing_gear_subscriber_ = nh.subscribe("landing_gear_contact_switches",
                                  100,
                                  &LandPlanner::processLandingGearMessage,
                                  this);
+    //Setting up the service client connection for disarm request
+    uav_arm_client_ = nh.serviceClient<iarc7_msgs::Arm>("uav_arm");
```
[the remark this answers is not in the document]
MiscM: wait so you want this above the landing gear subscriber?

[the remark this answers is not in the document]
amiller27: You can do it as part of the initializer list, which specifies how to construct each of the instance variables of the class.  What's happening how is `uav_arm_client_` is created by calling the default constructor on line 49, then that object is throw out and reassigned to the result of the function call on the right side of line 56.  What you can do instead is replace what's now line 49 with `uav_arm_client_(nh.serviceClient<iarc7_msgs::Arm>("uav_arm"))`

src/LandPlanner.cpp
MiscM: It just felt weird to have a post-impact action in a state called "brace for impact". I can take it out but I liked the symmetry it had with the takeoff controller and also I don't think run time is an issue once we've landed

src/LandPlanner.cpp
MiscM: wait are two tabs not equal to four spaces?

MiscM: nevermind figured it out

Riteshlogin: Sublime text has a setting that turns all of your tabs to spaces. I'm not sure if that's what you found when you figured it out, but if it isn't, it's what I recommend. I can show you on Sunday.

src/iarc7_motion/iarc_tasks/takeoff_task.py
aftersomemath: fc_comms is handling the angle mode stuff.

src/iarc7_motion/iarc_tasks/height_recovery_task.py
MiscM: the hit_roomba abstract worked in the sim when tested

src/TakeoffController.cpp
MiscM: fixed

PR #106, opened 2018-01-04
Fix the LLM safety response so that it actually occurs

aftersomemath:


src/LowLevelMotionController.cpp
```diff
@@ -409,7 +411,11 @@ int main(int argc, char **argv)
                 if(landPlanner.isDone())
                 {
                     ROS_INFO("Land completed");
-                    server.setSucceeded();
+
+                    if(!safety_client.isSafetyResponseActive())
```
amiller27: Is there a reason we only do this check here and not for the other states?

aftersomemath: Safety being active means that motion coordinator is down. The assumption was then made that nobody else would use the action interface (for simplicity) since this will all change soon. Landing doesn't return anything but success because anything else means the drone crashed (not a great system). So there was nothing else to check.

amiller27: Got it

PR #107, opened 2018-01-12
Support landing detector instead of contact switches

aftersomemath:
This doesn't touch the motion command coordinator which depends on landing gear switches for block and hit roomba.


aftersomemath:
Everything is tested

PR #111, opened 2018-02-01
Publishing P,I,D, values to thrust, pitch, and roll topics

MiscM:


include/iarc7_motion/PidController.hpp
```diff
@@ -35,7 +37,7 @@ class PidController
             const ros::Time& time,
             double& result,
             double derivative = std::numeric_limits<double>::quiet_NaN(),
-            bool log_debug=false);
+            bool log_debug=true);
```
Riteshlogin: So this makes log_debug true by default unless made false by whatever calls the function?

asaba96:
Was there anything more needed to be reviewed for this one before merging? It looked okay to me too.

PR #114, opened 2018-03-11
Rewrite the Acceleration Planner into a MotionPointInterpolator

aftersomemath:
Needs testing. PRing for visibility.

Before anyone says anything I did use git mv but git still decided to treat everything as a new file.


aftersomemath:
Tested, ready for merging once the PR for the UAVMotionPoint is figured out

https://github.com/Pitt-RAS/iarc7_msgs/pull/30

aftersomemath:
Ready for merging once Pitt-RAS/iarc7_msgs#30 is merged.

scripts/thrust_target_to_motion_point_transcriber.py
```diff
@@ -0,0 +1,27 @@
...
+    for twist in msg.twists:
+        motion_point = MotionPointStamped()
+        motion_point.header = twist.header
+        motion_point.motion_point.twist = twist.twist
+        motion_points.motion_points.append(motion_point)
+
+    motion_point_pub.publish(motion_points)
+
+
+if __name__ == '__main__':
```
asaba96: I really don't know how I feel about this node. I think it introduces a layer that can be easily removed by updating HLM to use the motion point message. Also, it can introduce lag between sending a request from HLM to LLM and idk if that is desirable or even acceptable.

asaba96: Also, considering that if this node were to crash, communication between HLM and LLM would be severed. This is obviously a bad situation so if this were to remain, we would need some kind of safety response. At minimum, it should be using `iarc7_safety` to ensure its bonded/running.

src/MotionPointInterpolator.cpp
```diff
@@ -0,0 +1,268 @@
...
+void MotionPointInterpolator::getTargetMotionPoint(
+                const ros::Time& current_time,
+                MotionPointStamped& target_motion_point)
+{
+
+    // Trim velocity queue when done we will have one or two velocities available.
+    trimMotionPointQueue(motion_point_targets_, current_time);
+
+    if(motion_point_targets_.size() == 1)
+    {
```
asaba96: Picky, but in our codebase, I see the `{` on the same line as the if statement and on the next line. Do we have a convention and can we adopt it universally? Personally I prefer same line.

src/MotionPointInterpolator.cpp
```diff
@@ -0,0 +1,268 @@
...
+    interpolated.motion_point.pose.position.x = interpolate(x, begin.motion_point.pose.position.x, end.motion_point.pose.position.x);
+    interpolated.motion_point.pose.position.y = interpolate(x, begin.motion_point.pose.position.y, end.motion_point.pose.position.y);
+    interpolated.motion_point.pose.position.z = interpolate(x, begin.motion_point.pose.position.z, end.motion_point.pose.position.z);
+
+    interpolated.motion_point.twist.linear.x = interpolate(x, begin.motion_point.twist.linear.x, end.motion_point.twist.linear.x);
+    interpolated.motion_point.twist.linear.y = interpolate(x, begin.motion_point.twist.linear.y, end.motion_point.twist.linear.y);
+    interpolated.motion_point.twist.linear.z = interpolate(x, begin.motion_point.twist.linear.z, end.motion_point.twist.linear.z);
+
+    // z interpolation is left in for now since
+    // there yaw is not supported by any other way in the controller
```
asaba96: Check your comment here for readability

asaba96:
This is also failing to build. Finally, have these changes been tested on the drone?

asaba96:
When I say drone I mean not just crazy fly. I don't know if it is a sufficient enough test platform for LLM changes.

scripts/thrust_target_to_motion_point_transcriber.py
aftersomemath: This node only exists because HLM hasnt been updated. I forsee a lot of discussion on that. How long till HLM is updated? If its a long time the safety response can be added but that requires changes to launch that I wanted to avoid.

This node is rather simple and has a very low chance of crashing.

src/MotionPointInterpolator.cpp
aftersomemath: The convention is next line but we decided to not pick on it code reviews and just fix it when we see it.

src/MotionPointInterpolator.cpp
```diff
@@ -0,0 +1,268 @@
...
+    // Set the current header stamp
+    interpolated.header.stamp = time;
+
+    return true;
+}
+
+// Interpolate between arbitrary values and waypoints
+// x is to be between 0-1 if the interpolation is to
+// range from start to end.
+double MotionPointInterpolator::interpolate(double x,
```
amiller27: Can we fix the formatting here?

scripts/thrust_target_to_motion_point_transcriber.py
```diff
@@ -0,0 +1,27 @@
...
+    rospy.init_node('thrust_target_to_motion_point_transcriber')
+
+    velocity_sub = rospy.Subscriber('movement_velocity_targets', TwistStampedArray, callback)
+    motion_point_pub = rospy.Publisher('motion_point_targets', MotionPointStampedArray, queue_size=0)
+
+    rate = rospy.Rate(30)
```
amiller27: Can we just replace this with `rospy.spin()`?

src/MotionPointInterpolator.cpp
amiller27: The convention we decided on is `{` on the same line, also convention says you need a space between `if` and `(`.  We're using the Google C++ style guide with some exceptions, but this isn't one of the exceptions.  However, I don't care, and I'd say Levi can fix it if he feels like it.
https://google.github.io/styleguide/cppguide.html#Conditionals

aftersomemath: Can't believe that I get this wrong so often.

src/MotionPointInterpolator.cpp
aftersomemath: Fixed

scripts/thrust_target_to_motion_point_transcriber.py
aftersomemath: Fixed

PR #118, opened 2018-03-21
Track roomba revamp

asaba96:
Track roomba now uses full PID. In addition, I made additional tweaks and enhancements to Track roomba, the task directory structure, and added comments to the QuadMove.action.


src/iarc7_motion/iarc_tasks/task_utilities/pid_controller.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python
+import rospy
+import numpy as np
+
+class PidSettings(object): 
+    def __init__(self, settings_dict):
+        self.kp = settings_dict.get('kp', 0)
```
amiller27: I don't think you want defaults here, you probably just want to use a normal dict lookup and let the `KeyError` get thrown if the setting isn't there.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -147,11 +154,24 @@ def get_desired_command(self):
...
+                y_diff = self._roomba_point.point.y + y_overshoot
+
+                x_success, x_vel_target = self._x_pid.update(x_diff, rospy.Time.now(), False)
+                y_success, y_vel_target = self._y_pid.update(y_diff, rospy.Time.now(), False)
+
+                # PID controller does setpoint - current; 
+                # the difference from do_transform_point is from the drone to the roomba, 
+                # which is the equivalent of current-setpoint, so take the negative response
+                if x_success: 
+                    x_vel_target = -x_vel_target + roomba_x_velocity
```
amiller27: You should use different names for the `x_vel_target` that you set earlier and this one here

src/iarc7_motion/iarc_tasks/task_utilities/pid_controller.py
```diff
@@ -0,0 +1,125 @@
...
+        if not np.isfinite(current_value):
+            rospy.logwarn(
+                    'Invalid argument to PidControllor.update (current_value = {}).'.format(current_value))
+            return False, response
+
+        difference = self._setpoint - current_value
+        p_term = self._p_gain * difference
+        response = p_term
+
+        if self._initialized:
```
amiller27: Why do we need to check this here? This should always be true if we get to this point

asaba96: Fixed

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: fixed

src/iarc7_motion/iarc_tasks/task_utilities/pid_controller.py
asaba96: fixed

PR #120, opened 2018-03-26
Use thrust model v2 on all platforms

aftersomemath:
Removes old platforms, breaks compatibility with V1.1. 1.1 will be further broken by the next PR involving the controller for height being changed to a second order controller instead of first order. All other platforms are working though.


aftersomemath:
Rebased

PR #121, opened 2018-03-27
Z velocity to position

aftersomemath:
@MiscM Started this and I wrapped up a few details.

Completely breaks functional compatibility but not software compatibility due to significant restructuring of HLM that is required. I plan on doing this restructuring as part of senior design so it can't be in this PR. The restructure will achieve linear profile interpolation (necessary for senior design) with a revamp of HLM that allows the intuitive nature of velocity and position targets to be used while still causing completely filled motion points to be output.

Technically the software is still correct, its just that LLM won't ever takeoff since the QuadVelocity controller looks at Z position which isn't being filled out by HLM.

### DIFF
--- a/scripts/waypoints_test.py
+++ b/scripts/waypoints_test.py
@@ -37,7 +37,7 @@ if __name__ == '__main__':
         except rospy.ServiceException as exc:
             print("Could not arm: " + str(exc))
 
-    # Target points in global (X, Y, Z) coordinates
+    # Target points in global (X, Y, Z, Yaw) coordinates
     waypoints = [
             (0, 0, .3, 0 * math.pi),
             (0, 0, .4, 0 * math.pi),
--- a/scripts/waypoints_test.py
+++ b/scripts/waypoints_test.py
@@ -39,8 +39,6 @@ if __name__ == '__main__':
 
     # Target points in global (X, Y, Z) coordinates
     waypoints = [
-            (0, 0, .1, 1 * math.pi),
-            (0, 0, .2, 0 * math.pi),
             (0, 0, .3, 0 * math.pi),
             (0, 0, .4, 0 * math.pi),
             (0, 0, .5, 0 * math.pi),
@@ -49,13 +47,12 @@ if __name__ == '__main__':
             (0, 0, .9, 1.5 * math.pi),
             (0, 0, .9, 1 * math.pi),
             (0, 0, .9, 1 * math.pi),
-            (0, 0, .7, 1 * math.pi),
-            (0, 0, .6, 1 * math.pi),
             (0, 0, .5, 1 * math.pi),
-            (0, 0, 0, 1.25 * math.pi),
+            (0, 0, .9, 1 * math.pi),
+            (0, 0, .5, 1 * math.pi),
             ]
     waypoints_iter = iter(waypoints)
-    target = next(waypoints_iter)
+    target = waypoints_iter.next()
 
     kP = 0.8
     gamma = 0.8
@@ -78,7 +75,7 @@ if __name__ == '__main__':
     rospy.logwarn("Takeoff success: {}".format(ground_interaction_client.get_result()))
 
     rate = rospy.Rate(30)
-    time = rospy.Time.now().to_sec() 
+    time = rospy.Time.now().to_sec()
     while not rospy.is_shutdown():
         try:
             (trans, rot) = tf_listener.lookupTransform('/map',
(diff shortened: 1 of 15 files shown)


amiller27:
So I'm thinking we wait to merge this until the rest of it is ready? Is that good with you?

aftersomemath:
Yes but please review so that we can fix things sooner rather than later.

scripts/waypoints_test.py
```diff
@@ -22,7 +21,7 @@ def constrain(x, l, h):
     safety_client = SafetyClient('motion_command_coordinator')
     assert safety_client.form_bond()
 
-    velocity_pub = rospy.Publisher('movement_velocity_targets', TwistStampedArray, queue_size=0)
+    motion_point_pub = rospy.Publisher('motion_point_targets', MotionPointStampedArray, queue_size=0)
```
Riteshlogin: What subscribes to this and are there any other PR's that this depends on?

scripts/waypoints_test.py
```diff
@@ -40,15 +39,20 @@ def constrain(x, l, h):
 
     # Target points in global (X, Y, Z) coordinates
```
Riteshlogin: This comment is not entirely accurate

scripts/waypoints_test.py
aftersomemath: Low level motion. Nothing depends on this but the whole PR depends on one that doesnt exist yet.

asaba96:
What kind of changes to HLM were you envisioning? Supporting motion points commands? Revamping the takeoff/land tasks?

asaba96:
Also please no breaking changes being merged. Otherwise you will force us to use an old commit version of master to do our work.

scripts/waypoints_test.py
asaba96: What does this pr depend on? Another topic or node? I thought you made a motion point transcriber node.

aftersomemath: It depends on a hlm outputting motion points with z position filled out.

scripts/waypoints_test.py
```diff
@@ -90,53 +94,43 @@ def constrain(x, l, h):
...
-            if abs(yaw_difference) < 0.15:
-                target = next(waypoints_iter, target)
+        motion_point.motion_point.pose.position.z = target[2]
+
+        motion_point_msg = MotionPointStampedArray()
+        motion_point_msg.motion_points = [motion_point]
+        motion_point_pub.publish(motion_point_msg)
+
+        #Sets next target after a set time has passed
+        if ((rospy.Time.now().to_sec() - time) > 1):
```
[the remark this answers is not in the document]
aftersomemath: It is never hit. This script just tests controller performance in a very raw manner.

scripts/waypoints_test.py
asaba96: Velocity commands require a z position now?

aftersomemath:
HLM needs to maintain a simple way to tasks to assign positions and velocities but needs to provide full motion point plans at all times. I plan to add a simple profile generator that generates plans from target positions or velocities.

scripts/waypoints_test.py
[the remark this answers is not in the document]
Riteshlogin: Hmm, ok. I think that by competition time, this should be replaced by logic that checks the status of the goal or the difference between the target and the current position instead.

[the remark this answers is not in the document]
Riteshlogin: Never mind

src/LandPlanner.cpp
```diff
@@ -100,28 +98,13 @@ bool LandPlanner::getTargetTwist(const ros::Time& time,
...
     }
 
-    if(state_ == LandState::ACCELERATE_TO_DESCENT_VELOCITY) {
-        requested_z_vel_ = std::max(requested_z_vel_ - ((time - last_update_time_).toSec() * descend_acceleration_), descend_rate_);
-        if(requested_z_vel_ <= descend_rate_) {
-            state_ = LandState::DESCEND;
-        }
-    }
-    else if(state_ == LandState::DESCEND)
+    if(state_ == LandState::DESCEND)
```
amiller27: Why are we not ramping the descent rate anymore?

aftersomemath: We are still ramping the setpoint, its just that I didn't want to handle ramping 2 derivatives down. In theory this should be replaced by a plan generator in HLM. Which may happen. This gave acceptable performance in the sim.

scripts/waypoints_test.py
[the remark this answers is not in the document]
aftersomemath: Fixed

scripts/waypoints_test.py
aftersomemath: Fixed

src/LandPlanner.cpp
aftersomemath: Done

PR #122, opened 2018-03-28
Linear Motion Profile Generator

aftersomemath:
Currently supports generating motion profiles from velocity commands. Produces completely filled motion point arrays suitable for use for feedforward and setpoint type switching in LLM.

Fixes the incompatibilities introduced by #121 .

Personally, I'm a big fan of this approach for generating motion profiles for tasks in HLM. It allows plans to be specified with little effort while respecting the limits of the controller.


src/iarc7_motion/task_command_handler.py
```diff
@@ -250,23 +264,15 @@ def handle_ground_interaction_done(self, status, result):
...
-    Sends twist (x, y, z, and angular velocities) to LLM
+    Sends motion point stamped array to LLM
 
     Args:
-        twist: TwistStampedArray (array of twists) or just a single twist stamped.
+        motion_point_stamped_array: MotionPointStampedArray
     """
-    def _publish_twist(self, twist):
-        if type(twist) is TwistStampedArray:
-            self._last_twist = twist.twists[-1]
```
amiller27: Let's rip out `self._last_twist` entirely if we don't use it anymore

amiller27: And rip it out of the state monitor entirely

src/iarc7_motion/linear_motion_profile_generator.py
```diff
@@ -0,0 +1,209 @@
...
+
+        # Initialize the velocities array with the starting velocity
+        velocities = [v_start]
+
+        # Generate velocities corresponding to the acceleration period
+        if accel_steps > 0:
+            accel_times = (np.linspace(0.0, accel_steps*self._PROFILE_TIMESTEP, accel_steps)
+                     * accel_steps
+                     * self._PROFILE_TIMESTEP)
+            velocities.extend([a_target * accel_time + v_start for accel_time in accel_times])
```
amiller27: This is going to add two timesteps at `v_start`

src/iarc7_motion/linear_motion_profile_generator.py
```diff
@@ -0,0 +1,209 @@
...
+        np_to_msg(p_start, motion_point.motion_point.pose.position)
+
+        plan.motion_points.append(motion_point)
+        pose = PoseStamped()
+        pose.header.stamp = motion_point.header.stamp
+        pose.pose = motion_point.motion_point.pose
+        pose_only_plan.poses.append(pose)
+
+        # Generate the motion profile for all the remaining velocities and accelerations
+        for i in range(1, velocities.shape[0]-1):
```
amiller27: Is there a reason we're throwing out the last point in `velocities`?

aftersomemath: Yes because we don't have an acceleration for the last time.

src/iarc7_motion/task_command_handler.py
aftersomemath: Agreed

src/iarc7_motion/linear_motion_profile_generator.py
```diff
@@ -0,0 +1,209 @@
...
+
+            # Check that the time is at least equal to the current time
+            # before generating a plan
+            if start_motion_point.header.stamp < rospy.Time.now():
+                start_motion_point.header.stamp = rospy.Time.now()
+            return start_motion_point
+
+        else:
+            # Make sure a starting point newer than the last sent time is sent
+            for i in range(1, len(self._last_motion_plan.motion_points)):
```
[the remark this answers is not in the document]
asaba96: Do we assume LLM has started executing the motion points and we want to interpolate from the absolute first guaranteed one that hasn't been executed?

src/iarc7_motion/linear_motion_profile_generator.py
```diff
@@ -0,0 +1,209 @@
...
+        v_delta = v_desired - v_start
+
+        # Assign a direction to the target acceleration
+        a_target = self._TARGET_ACCEL * v_delta / np.linalg.norm(v_delta)
+
+        # Find the time required to accelerate to the desired velocity
+        acceleration_time = min(np.linalg.norm(v_delta) / self._TARGET_ACCEL, self._PLAN_DURATION)
+
+        # Use the rest of the profile duration to hold the velocity
+        steady_velocity_time = self._PLAN_DURATION - acceleration_time
```
asaba96: What happens in the case that `self._PLAN_DURATION` is smaller than the `acceleration_time`?

amiller27: It can't be, because of line 139

asaba96: Wow i shouldn’t code review at 1am in the morning

aftersomemath: Well I wrote this around 3-4am

src/iarc7_motion/linear_motion_profile_generator.py
aftersomemath: Fixed

src/iarc7_motion/task_command_handler.py
aftersomemath: Can't be ripped out but added a fix that is in the spirit of the way it used to be.

src/iarc7_motion/linear_motion_profile_generator.py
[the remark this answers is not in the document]
aftersomemath: The motion point between the point just before the current time and just after the current time is interpolated and assumed to be the current state of the drone. The motion plan is then re-generated from that state.

src/iarc7_motion/linear_motion_profile_generator.py
```diff
@@ -0,0 +1,209 @@
...
+            # Check that the time is at least equal to the current time
+            # before generating a plan
+            if start_motion_point.header.stamp < rospy.Time.now():
+                start_motion_point.header.stamp = rospy.Time.now()
+            return start_motion_point
+
+        else:
+            # Make sure a starting point newer than the last sent time is sent
+            for i in range(1, len(self._last_motion_plan.motion_points)):
+                if self._last_motion_plan.motion_points[i].header.stamp > time:
```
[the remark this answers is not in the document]
aftersomemath: Because the first point has to be a point newer than the current time for the indexing to work.

src/iarc7_motion/task_command_handler.py
aftersomemath: Some more was removed as discussed.

PR #123, opened 2018-03-29
Go to roomba

Quinnan-Gill:
This is a new file in order that uses the translate stop planner to move to a roomba making it them easier to use track roomba


src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+                if(transStamped.transform.translation.z > self._MIN_MANEUVER_HEIGHT):
+                    """ quarantine"""
+                    # check if the roomba is in range
+                    if not self._check_roomba_in_sight():
+                        return (TaskAborted(msg='The provided roomba is not in sight of quad'),)
+                    elif self._on_ground():
+                        return (TaskDone(),)
+
+                    try:
+                        roomba_transform = self.topic_buffer.get_tf_buffer().lookup_transform(
```
asaba96: Same comment as below. Why do we need to lookup any transforms? Why can’t we just use the current roomba position in ‘self._roomba_odometry‘? Is it to guarantee the appropriate frame?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+                        return (TaskDone(), VelocityCommand(hold_twist))
+                    elif not self._path_holder.is_done():
+                        return (TaskRunning(), VelocityCommand(hold_twist))
+                    else:
+                        return (TaskDone(), VelocityCommand(hold_twist))
+                else:
+                    return (TaskFailed(msg='Fell below minimum manuever height during translation'),)
+
+
+            return (TaskAborted(msg='Impossible state in takeoff task reached'))
```
asaba96: Comment needs updated

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+    translate = 1
+
+class GoToRoombaTask(AbstractTask):
+
+    def __init__(self, task_request):
+        super(GoToRoombaTask, self).__init__()
+
+        self._roomba_id = task_request.frame_id + '/base_link'
+
+        if self._roomba_id is None:
```
asaba96: This check will always pass; rather it needs to check to see if roomba ID is equal to ‘/base_link‘

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+            raise ValueError('A null roomba id was provided')
+
+        self._roomba_odometry = None
+        self._roomba_point = None
+        self._limiter = AccelerationLimiter()
+
```
asaba96: You create class variables, such as roomba point check level quad, but don’t initiate them here in the constructor.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+        self._canceled = False;
+        self._last_update_time = None
+        self._current_velocity = None
+        self._transition = None
+
+        self._lock = threading.RLock()
+
+        self._x_position = task_request.x_position
```
asaba96: The task request wouldn’t be sending in a x or y position, but rather a roomba id, so these are going to be 0.0. This won’t  impact the translate stop planner, but it’s unnecessary code. You can just pass in None for the x and y initial positions to the translate stop planner and that will also work.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+        self._y_position = task_request.y_position
+
+
+        try:
+            self._TRANSLATION_XYZ_TOLERANCE = rospy.get_param('~translation_xyz_tolerance')
+            self._TRANSFORM_TIMEOUT = rospy.get_param('~transform_timeout')
```
asaba96: There’s a number of parameters here that get retrieved and never used.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+            self._MIN_MANEUVER_HEIGHT = rospy.get_param('~min_maneuver_height')
+            self._MAX_HORIZ_SPEED = rospy.get_param('~max_translation_speed')
+            self._MAX_TASK_DIST = rospy.get_param('~max_roomba_dist')
+            self._K_X = rospy.get_param('~k_term_tracking_x')
+            self._K_Y = rospy.get_param('~k_term_tracking_y')
+            self._z_position = rospy.get_param('~track_roomba_height')
+        except KeyError as e:
+            rospy.logerr('Could not lookup a parameter for xyztranslation task')
```
asaba96: Comment needs updated.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+            raise
+
+        # Check that we aren't being requested to go below the minimum maneuver height
+        # Error straight out if that's the case. If we are currently below the minimum height
+        # It will be caught and handled on the next update
+        if self._z_position < self._MIN_MANEUVER_HEIGHT :
+            raise ValueError('Requested z height was below the minimum maneuver height')
+
+        # fc status?
```
asaba96: What does this comment mean?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+import rospy
+import tf2_ros
+import tf2_geometry_msgs
+import threading
+
+from geometry_msgs.msg import TwistStamped
+from geometry_msgs.msg import Point
+from geometry_msgs.msg import PointStamped
+
+from task_utilities.acceleration_limiter import AccelerationLimiter
```
asaba96: This import isn’t needed.

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
```diff
@@ -52,6 +51,12 @@ def get_xyz_hold_response(self):
                 response.header.frame_id = 'level_quad'
             return response
 
+    def reinitialize_translation_stop_planner(self, x=None, y=None, z=None):
```
asaba96: We should do checks here that the new values aren’t None before reassignment. Otherwise we could have undefined behavior or unhandled exceptions.

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
```diff
@@ -106,14 +111,8 @@ def _calculate_triangle_acceleration(self,
             # otherwise just keep accelerating
             if required_deceleration < self._desired_acceleration:
                 # If these two things are true we've been decelerating too fast
-                if (self._acceleration_went_negative and
```
asaba96: Can someone explain why this was removed and why it isn’t needed anymore?

param/low_level_motion_sim.yaml
```diff
@@ -52,7 +52,7 @@ level_flight_required_height: 0.35
 level_flight_required_hysteresis: 0.10
 
 update_frequency: 60.0
-startup_timeout: 10.0
+startup_timeout: 20.0
```
asaba96: Is there a reason we changed this?

asaba96:
(Making a comment) I am curious as to the reasoning behind using transforms for everything when we can get the roombas map positions from the roomba odometry. The same can be said for the drone’s current height. Finally, if we had an ending radius specified by the task request (which we probably will want anyway) and provided that to the translate stop planner, we won’t need to find how far away the roomba is from the drone, since the translate stop planner will be doing this check already.

src/iarc7_motion/iarc_task_action_client.py
```diff
@@ -0,0 +1,59 @@
+
+import threading
```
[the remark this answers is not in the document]
Quinnan-Gill: Sorry this response took so long, got VMware fusion and I was having trouble trying to move the changes to that.

As For the file. I delete it, it was from the last project I was working on.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+                                                    self._z_position)
+
+        self._state = GoToRoombaState.init
+
+
+    def get_desired_command(self):
+        with self._lock:
+            if self._canceled:
+                return (TaskCanceled(),)
+            if self._state == GoToRoombaState.init:
```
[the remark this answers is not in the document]
Quinnan-Gill: Fixed

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+                    rospy.logerr('GoToRoombaTask: Exception when looking up transform')
+                    rospy.logerr(ex.message)
+                    return (TaskAborted(msg = 'Exception when looking up transform during go_to_roomba'),)
+
+                if(transStamped.transform.translation.z > self._MIN_MANEUVER_HEIGHT):
+                    """ quarantine"""
+                    # check if the roomba is in range
+                    if not self._check_roomba_in_sight():
+                        return (TaskAborted(msg='The provided roomba is not in sight of quad'),)
+                    elif self._on_ground():
```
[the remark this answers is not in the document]
Quinnan-Gill: Removed this

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
Quinnan-Gill: I added it a while ago, I don't remember what it does, so I remove it

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
Quinnan-Gill: Removed the import

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
Quinnan-Gill: Added the check

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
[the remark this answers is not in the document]
Quinnan-Gill: Actually it is used to initialize the value for roomba odometer

param/low_level_motion_sim.yaml
amiller27: This is just so that stuff doesn't time out in VMs, it shouldn't be in the PR

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+                                                            stamped_point, roomba_check_pos)
+
+
+                    roomba_x_velocity = self._roomba_odometry.twist.twist.linear.x
+                    roomba_y_velocity = self._roomba_odometry.twist.twist.linear.y
+
+                    x_vel_target = self._roomba_point_map.point.x + roomba_x_velocity
+                    y_vel_target = self._roomba_point_map.point.y + roomba_y_velocity
+                    
+                    self._path_holder.reinitialize_translation_stop_planner(x_vel_target,
```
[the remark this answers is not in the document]
Quinnan-Gill: Wouldn't this mean that the make it so that it could track a moving Roomba's position. I also played around with this quite a bit and this seemed to be the combination that worked

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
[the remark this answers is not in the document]
asaba96: no, I was referring to the `self._on_ground()` call, which we shouldn't need to be checking.

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
```diff
@@ -53,6 +53,10 @@ def get_xyz_hold_response(self):
 
     def reinitialize_translation_stop_planner(self, x=None, y=None, z=None):
         with self._lock:
+            if (x is None or y is None or z is None):
+                rospy.logerr('Exception: Undefined behavoir when retinitializing translate stop planner')
```
asaba96: I was thinking of if you do not change the value (i.e. you pass in `None` for x or leave the default value), then the function will retain the current end goal for the respective dimension (x in my example).

param/low_level_motion_sim.yaml
asaba96: So please revert this value back.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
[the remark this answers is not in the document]
asaba96: Can you please clarify your remark?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,186 @@
...
+                self._roomba_odometry = odometry
+                return True
+        return False
+
+    def _check_max_roomba_range(self,):
+        _distance_to_roomba = math.sqrt(self._roomba_point_level_quad.point.x**2 +
+                            self._roomba_point_level_quad.point.y**2)
+        return (_distance_to_roomba <= self._MAX_TASK_DIST * 0.75)
+
+    def _on_ground(self):
```
asaba96: We should not be checking if we are on the ground in this task, as it shouldn't be landing the drone.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,186 @@
...
+        return self.topic_buffer.get_landing_message().data
+    """ quarantive methods end """
```
asaba96: Quarantine comment

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,186 @@
...
+                        return (TaskRunning(), VelocityCommand(hold_twist))
+                    else:
+                        return (TaskDone(), VelocityCommand(hold_twist))
+                else:
+                    return (TaskFailed(msg='Fell below minimum manuever height during translation'),)
+
+
+            return (TaskAborted(msg='Impossible state in takeoff task reached'))
+
+    """ qurantine methods """
```
asaba96: Quarantine comment

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+
+                    x_vel_target = self._roomba_point_map.point.x + roomba_x_velocity
+                    y_vel_target = self._roomba_point_map.point.y + roomba_y_velocity
+                    
+                    self._path_holder.reinitialize_translation_stop_planner(x_vel_target,
+                                                                            y_vel_target,
+                                                                            self._z_position)
+
+
+                    """ quarantine end"""
```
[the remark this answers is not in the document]
Quinnan-Gill: Removed the comment

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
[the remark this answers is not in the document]
Quinnan-Gill: go to Roomba will not work if I just do the Roomba odometery (this meaning that I just do `self._roomba_odometry.twist.twist.linear.x`)

asaba96:
Also forgot to say, this needs rebased

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
Quinnan-Gill: Removed

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
Quinnan-Gill: Removed the comment

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
Quinnan-Gill: Removed the comment

param/low_level_motion_sim.yaml
Quinnan-Gill: Change is reverted

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
Quinnan-Gill: Ok so don't raise an error?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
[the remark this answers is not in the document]
Quinnan-Gill: Actually is there another property of self._roomba_odometry that will do what is above but require no transforms?

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+                                        rospy.Duration(self._TRANSFORM_TIMEOUT))
+                        roomba_check_pos = self.topic_buffer.get_tf_buffer().lookup_transform(
+                                         'level_quad',
+                                         self._roomba_id,
+                                         rospy.Time(0),
+                                         rospy.Duration(self._TRANSFORM_TIMEOUT))
+                    except (tf2_ros.LookupException,
+                            tf2_ros.ConnectivityException,
+                            tf2_ros.ExtrapolationException) as ex:
+                        rospy.logerr('ObjectTrackTask: Exception when looking up transform')
```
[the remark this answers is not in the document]
Quinnan-Gill: Changed the comment

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
Quinnan-Gill: Comment changed

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
Quinnan-Gill: Changed the statement to check if it is equal to `'/base_link'`

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
Quinnan-Gill: Changed the values to None

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
asaba96: No, but merely check each value and see if its None. If it is, then don't change the goal for that value in the planner.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,190 @@
...
+                    y_vel_target = self._roomba_point_map.point.y + roomba_y_velocity
+                    
+                    self._path_holder.reinitialize_translation_stop_planner(x_vel_target,
+                                                                            y_vel_target,
+                                                                            self._z_position)
+
+
+                    """ quarantine end"""
+                    hold_twist = self._path_holder.get_xyz_hold_response()
+                    if self._check_max_roomba_range():
```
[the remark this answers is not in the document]
Quinnan-Gill: I see your point, but I do not know how to do this.

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
```diff
@@ -52,6 +51,15 @@ def get_xyz_hold_response(self):
...
             return response
 
+    def reinitialize_translation_stop_planner(self, x=None, y=None, z=None):
+        with self._lock:
+            if not x is None:
+                self._hold_x = x
+            if not y is None:
+                self._hold_y = y
+            if not z is None:
+                self._hold_z
```
asaba96: You’re missing the assignment here

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -83,6 +83,7 @@ def __init__(self, task_request):
 
     def get_desired_command(self):
         with self._lock:
+
```
asaba96: Possible to remove the cosmetic line addition? Otherwise I think you might run into merge issues when you rebase.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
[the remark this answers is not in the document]
asaba96: For clarification, I was imagining a check like this: https://github.com/Pitt-RAS/iarc7_motion/blob/master/src/iarc7_motion/iarc_tasks/track_roomba_task.py#L98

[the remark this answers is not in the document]
Quinnan-Gill: This is already done on the line below (line 99)

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
Quinnan-Gill: Fixed

src/iarc7_motion/iarc_tasks/track_roomba_task.py
Quinnan-Gill: Removed that line

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
Quinnan-Gill: I don't remember why I removed it

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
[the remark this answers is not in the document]
Quinnan-Gill: Added the check

Quinnan-Gill:
Also, you were write in not needing all those transforms. I all of the ones that I thought were not needed.

asaba96:
This needs rebased. Additionally, I think it would be nice for the task request to specify an ending radius to the roomba. This would mean that the translate stop planner would accept a customizable radius instead of loading it in as a parameter. @amiller27 any thoughts on this functionality?

asaba96:
Specifically I’m referring to this in the translate stop planner: https://github.com/Pitt-RAS/iarc7_motion/blob/master/src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py#L27

src/iarc7_motion/iarc_tasks/track_roomba_task.py
Quinnan-Gill: Fixed this.

src/iarc7_motion/iarc_tasks/go_to_roomba_task.py
```diff
@@ -0,0 +1,140 @@
...
+        self._lock = threading.RLock()
+
+        self._x_position = None
+        self._y_position = None
+
+
+        try:
+            self._TRANSFORM_TIMEOUT = rospy.get_param('~transform_timeout')
+            self._MIN_MANEUVER_HEIGHT = rospy.get_param('~min_maneuver_height')
+            self._MAX_TASK_DIST = rospy.get_param('~max_roomba_dist')
```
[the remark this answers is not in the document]
Quinnan-Gill: Deleted

src/iarc7_motion/iarc_tasks/track_roomba_task.py
amiller27: For future reference, removing changes like this will not make the rebase easier, if anything you'll have to fix the merge twice as many times (unless, to remove the change, you amend the commit instead of adding a new commit).

asaba96: Yeah he’d have to squash his commits.

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
asaba96: @amiller27 @aftersomemath do either of you know why?

amiller27:
That functionality is out of scope for this PR.

src/iarc7_motion/iarc_tasks/task_utilities/translate_stop_planner.py
amiller27: I said to remove it because it doesn't really make sense for a changing endpoint.  This is fine for now; we might need some extra stuff later if we have problems with oscillations near the endpoint, but that's out of scope.

asaba96:
@Quinnan-Gill I rebased this and added in the ending radius functionality that I needed for my work in iarc7_abstract. @amiller27 @aftersomemath this should be good for final review.

PR #124, opened 2018-04-02
Feed forward for height

aftersomemath:
Use feed forward for height hold which significantly increases performance.

This revealed scary bugs in the thrust model, linear motion profile generator, and some minor issues with the land planner. These were fixed.

Fixing these issues solved the height oscillation in the simulator and the crazyflie. Adding feed forward significantly reduced lag.


aftersomemath:
Can this be merged then? Testing has revealed a great performance increase.

asaba96:
Can I review first?

PR #126, opened 2018-04-08
Local obstacle avoider

4ndr3w:
This PR adds an obstacle avoider to the motion command coordinator and as a task helper. The obstacle avoider identifies obstacles by sweeping a vector around the drone (starting from the requested velocity vector) and computing the norm of the obstacle's projection onto the flight vector. The task helper will pick the first flight vector that has no obstacles violating the distance limits. The resulting flight vector will have the same magnitude as the requested flight vector.

The "idle" obstacle avoider runs inside the motion command coordinator and replaces setting zero velocity when no task is running. If no obstacles are within a certain radius, zero velocity will be commanded. Otherwise, the obstacle avoider will drive the drone in the direction of the region that has the fewest obstacles.

This PR also implements the obstacle avoider task helper inside `velocity_task`

### DIFF
--- a/param/obstacle_avoider.yaml
+++ b/param/obstacle_avoider.yaml
@@ -1,0 +1,12 @@
+# The maximum perpendicular distance between a velocity vector and an obstacle before the obstacle is ignored
+obst_avoid_norm_limit: 1
+# How many steps to take when searching for a safe flight vector (effectivly determining the resolution)
+# Steps are sampled over a range of [0, 2*pi]
+obst_avoid_step_size: 90
+# Obstacles outside this radius are ignored
+obst_avoid_radius: 1
+# Flight vector magnitude - only used by the idle avoider
+obst_avoid_magnitude: 0.2
+# Minimum avoid flight vector magnitude - only used by the task helper
+obst_avoid_min_magnitude: 0.1
+
--- a/src/iarc7_motion/iarc_tasks/task_utilities/obstacle_avoid_helper.py
+++ b/src/iarc7_motion/iarc_tasks/task_utilities/obstacle_avoid_helper.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python
+import rospy
+import threading
+import tf2_geometry_msgs
+import tf2_ros
+import numpy as np
+
+from math import sin, cos, atan2, pi
+
+from iarc7_safety.iarc_safety_exception import IARCSafetyException
+from iarc7_msgs.msg import ObstacleArray
+from geometry_msgs.msg import PointStamped
+
+class ObstacleAvoider(object):
+    def __init__(self, tf_buffer):
+        self._lock = threading.RLock()
+
+        # Maximum allowed distance from an obstacle to its projection on the nearest flight vector
+        self._unsafe_obstacle_threshold = rospy.get_param("~obst_avoid_norm_limit")
+        # Obstacles outside this radius are ignored
+        self._obstacle_radius = rospy.get_param("~obst_avoid_radius")
+        # Step size used when searching for alternative vectors when the requested vector is unsafe
+        self._vector_step_size = rospy.get_param("~obst_avoid_step_size") # Divide by 180 to get degrees/step
+        # Minimum avoid vector
+        self._minimum_magnitude = rospy.get_param("~obst_avoid_min_magnitude")
+        # Transform timeout
+        self._timeout = rospy.Duration(rospy.get_param("~transform_timeout"))
+        self._obstacle_points = None
+
+        self._tf_buffer = tf_buffer
+
+        with self._lock:
+            self._obstacle_subscriber = rospy.Subscriber("/obstacles", ObstacleArray, self._update_obstacles)
+
+    def wait_until_ready(self, startup_timeout):
+        while (rospy.Time.now() == rospy.Time(0) and not rospy.is_shutdown()):
+            # Wait for time to be initialized
+            rospy.sleep(0.005)
+        if rospy.is_shutdown():
+            raise rospy.ROSInterruptException()
+
+        start_time = rospy.Time.now()
+
+        while ((self._obstacle_points is None)
+               and rospy.Time.now() < start_time + startup_timeout
+               and not rospy.is_shutdown()):
+            rospy.sleep(0.005)
+        if rospy.Time.now() >= start_time + startup_timeout:
+            raise IARCFatalSafetyException('SafetyMonitor timed out on startup')
+        if rospy.is_shutdown():
+            raise rospy.ROSInterruptException()
+
+    def _update_obstacles(self, obstacles):
+        with self._lock:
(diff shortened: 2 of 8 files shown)


src/iarc7_motion/idle_obstacle_avoider.py
```diff
@@ -0,0 +1,97 @@
...
+                if vector_is_safe and startOpenRegion is None:
+                    startOpenRegion = angle
+                elif not vector_is_safe and startOpenRegion is not None:
+                    regions.append([startOpenRegion, angle])
+                    startOpenRegion = None
+
+            # Handle ending on a safe region
+            if startOpenRegion is not None:
+                regions.append([startOpenRegion, 2*pi])
+
```
amiller27: I think we also want to merge regions like `[a, 2pi]` and `[0, b]` or otherwise handle that case

src/iarc7_motion/idle_obstacle_avoider.py
```diff
@@ -0,0 +1,97 @@
...
+            for r in regions:
+                if r[1]-r[0] > biggestRegion:
+                    biggestRegion = r[1]-r[0]
+                    region = r
+
+
+            # All regions are safe
+            if np.isclose(region[0], 0) and np.isclose(region[1], 2*pi):
+                return np.array([0,0,0])
+            elif region is None:
```
amiller27: This check needs to come before line 88

src/iarc7_motion/iarc_tasks/task_utilities/obstacle_avoid_helper.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python
```
amiller27: Same comments as for idle obstacle avoider

amiller27: Also can we make this a part of the task topic buffer in AbstractTask so we don't need to create a new subscriber for every task?

asaba96:
Are we assuming the tasks are going to handle getting safe velocities on their own? Because we have not trusted them in the past and I didnt see anything here that maintains that lack of trust. For example, the idle obstacle avoidance only runs when there isn’t a task running. So are we okay with trusting the tasks to not take the drone into a dangerous path?

src/iarc7_motion/iarc_tasks/task_utilities/obstacle_avoid_helper.py
```diff
@@ -0,0 +1,84 @@
...
+import rospy
+from iarc7_msgs.msg import ObstacleArray
+import threading
+from math import sin, cos, atan2, pi
+import tf2_geometry_msgs
+import tf2_ros
+from geometry_msgs.msg import PointStamped
+import numpy as np
+
+class ObstacleAvoider:
```
asaba96: This should inherit from `object`

amiller27:
@asaba96 That's out of scope for this PR

src/iarc7_motion/idle_obstacle_avoider.py
```diff
@@ -0,0 +1,97 @@
...
+import rospy
+from iarc7_msgs.msg import ObstacleArray
+import threading
+from math import sin, cos, atan2, pi
+import tf2_geometry_msgs
+import tf2_ros
+from geometry_msgs.msg import PointStamped
+import numpy as np
+
+class IdleObstacleAvoider:
```
asaba96: Same as obstacle avoider, inherit from `object`

param/obstacle_avoider.yaml
```diff
@@ -0,0 +1,12 @@
+# How far an obstacle must be from a flight vector for it to be considered safe
```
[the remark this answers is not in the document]
4ndr3w: Yep, I'll update the comment to make it clearer

param/obstacle_avoider.yaml
```diff
@@ -0,0 +1,12 @@
...
+obst_avoid_norm_limit: 1
```
Riteshlogin: Does this mean that you only consider continuing to move forward or turning around when avoiding obstacles? I assume only 1 step in sampling between 0 - 2pi results in just angles of 0 and 2pi off the current velocity vector being considered.

4ndr3w: As Aaron commented above, this is the farthest perpendicular distance in meters that an obstacle can be to a velocity vector for it to be discarded. I'm updating the comment to make it clearer.

I'm not sure what you mean by continuing forward or turning around - in practice, if the obstacle avoider is given a constant forward velocity vector toward an obstacle, the drone will translate or "slide around" the obstacle.

Riteshlogin: Aaron's comment made the first comment in that file not show up on my screen, so I assumed that the "How many steps to take" comment corresponded to the "obst_avoid_norm_limit" variable. Obviously that makes no sense now, but I thought limit meant the limit on the number of steps

asaba96:
I have no issues with this once Aaron's comment above is addressed.

src/iarc7_motion/idle_obstacle_avoider.py
```diff
@@ -34,22 +36,24 @@ def _update_obstacles(self, obstacles):
...
+                    tf2_ros.ConnectivityException,
+                    tf2_ros.ExtrapolationException) as ex:
+                rospy.logwarn("ObstacleAvoider Couldn't lookup transform from {} to level_quad".format(obstacles.header.frame_id))
 
     def get_safest(self, magnitude):
         with self._lock:
             regions = []
             startOpenRegion = None
-            for angle in np.linspace(0, 2*pi, num=self._vector_step_size):
+            for angle in np.linspace(0, 2*pi, num=self._vector_step_size*2):
```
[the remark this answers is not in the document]
asaba96: @4ndr3w why the 2?

amiller27:
Can you integrate this into the safety system like I did in #134 ?

4ndr3w:
@amiller27 Yep!

asaba96:
Is this ready?

src/iarc7_motion/iarc_tasks/task_utilities/obstacle_avoid_helper.py
```diff
@@ -0,0 +1,105 @@
...
+            raise rospy.ROSInterruptException()
+
+        start_time = rospy.Time.now()
+
+        while ((self._obstacle_points is None)
+               and rospy.Time.now() < start_time + startup_timeout
+               and not rospy.is_shutdown()):
+            rospy.sleep(0.005)
+        if rospy.Time.now() >= start_time + startup_timeout:
+            raise IARCFatalSafetyException('SafetyMonitor timed out on startup')
```
asaba96: I don't see the IARC Safety Exception imported in this file.

aftersomemath:
Just needs rebase and the import.

PR #130, opened 2018-05-16
Use PID on velocity and P term on position. Use dynamic reconfigure for PID parameters.

aftersomemath:
V1.9 is currently not supported.

Loading the defaults for a platform from rosparam server and overwriting the dynamic reconfigure parameters is not necessary and I tested that it works. https://answers.ros.org/question/28327/dynamic-reconfigure-default-parameters/?answer=28332#post-id-28332


aftersomemath:
V1.9 is gone, V2.0 is in here and tuned. Should be good to go.

param/thrust_models/thrust_model_2.0.yaml
```diff
@@ -1,6 +1,6 @@
 # Generated on 2018-05-18 21:21:38.749113
 
-response_lag: 0.03184030109665681
+response_lag: 0.0 #0.03184030109665681
```
asaba96: whats going on with this?

src/TakeoffController.cpp
```diff
@@ -108,14 +108,14 @@ bool TakeoffController::update(const ros::Time& time,
...
-
+    bool success = true;
     // Get the current transform (xyz) of the quad
-    geometry_msgs::TransformStamped transform;
-    bool success = transform_wrapper_.getTransformAtTime(transform,
-                                                         "map",
-                                                         "base_footprint",
-                                                         time,
-                                                         update_timeout_);
+    //geometry_msgs::TransformStamped transform;
```
asaba96: why is this all commented out?

param/thrust_models/thrust_model_2.0.yaml
aftersomemath: We can't account for response lag because state estimation is not extrapolated into the future.

src/QuadVelocityController.cpp
```diff
@@ -462,10 +462,24 @@ double QuadVelocityController::yawFromQuaternion(
...
 }
 
+void QuadVelocityController::commandForAccel(
+        const Eigen::Vector3d& accel,
+        double& pitch,
+        double& roll,
+        double& thrust)
+{
+    const auto a_hat = accel.normalized();
+    roll = -std::asin(a_hat(1));
```
aftersomemath: Should roll be calculated with a_hat_no_pitch similar to how a_hat_no_roll is to calculate pitch? The total thrust vector that is being projected onto the y axis shouldn't depend on the thrust in the x axis.

src/TakeoffController.cpp
aftersomemath: It needed to be completely removed actually which I just did.

src/QuadVelocityController.cpp
amiller27: No, because of the order in which they're applied you don't need it for roll

include/iarc7_motion/PidController.hpp
```diff
@@ -42,19 +40,19 @@ class PidController
     void resetAccumulator();
 
 private:
-    const double p_gain_;
-    const double i_gain_;
-    const double d_gain_;
+    double& p_gain_;
```
amiller27: Why do these have to be references?

src/LowLevelMotionController.cpp
```diff
@@ -156,7 +170,8 @@ int main(int argc, char **argv)
 
     ros::Rate limit_check_for_simulated_time = ros::Rate(30);
     // Wait for a valid time in case we are using simulated time (not wall time)
-    while (ros::ok() && ros::Time::now() == ros::Time(0)) {
+    // Also wait for dynamic reconfigure to be called once
+    while ((ros::ok() && ros::Time::now() == ros::Time(0)) || !dynamic_reconfigure_called) {
```
amiller27: Don't you want the ros ok to be anded with the or of the other two here?

src/QuadVelocityController.cpp
```diff
@@ -248,80 +231,88 @@ bool QuadVelocityController::update(const ros::Time& time,
...
             return false;
         }
     }
 
     // Fill in the uav_command's information
     uav_command.header.stamp = time;
 
-    double hover_accel = 9.8;
-    double tilt_accel = 0.0;
+    // TODO add accel setpoints from plan here
```
amiller27: We can delete this comment

src/QuadVelocityController.cpp
```diff
@@ -429,22 +427,24 @@ bool QuadVelocityController::waitUntilReady()
     return true;
 }
 
-void QuadVelocityController::updatePidSetpoints(double current_yaw)
+void QuadVelocityController::updatePidSetpoints(double current_yaw, Eigen::VectorXd& odometry)
```
amiller27: Let's pass this by const&

src/QuadVelocityController.cpp
aftersomemath: As discussed I was mistakenly applying rotation in the global frame instead of the quad frame after each rotation.

include/iarc7_motion/PidController.hpp
aftersomemath: By being references the actual memory locations are the same as the array of settings the is passed in and controlled by the dynamic reconfigure callback. It's poor practice but I don't think its worth fixing with a settings object. I'll do it you insist though.

src/QuadVelocityController.cpp
aftersomemath: Fixed

src/iarc7_motion/iarc_tasks/track_roomba_task.py
```diff
@@ -146,9 +146,8 @@ def get_desired_command(self):
                 if not self._check_max_roomba_range():
                     return (TaskAborted(msg='The provided roomba is not found or not close enough to the quad'),)
 
-                roomba_x_velocity = self._roomba_odometry.twist.twist.linear.x
-                roomba_y_velocity = self._roomba_odometry.twist.twist.linear.y
-                roomba_velocity = math.sqrt(roomba_x_velocity**2 + roomba_y_velocity**2)
+                roomba_x_velocity = 0.0 #self._roomba_odometry.twist.twist.linear.x
```
aftersomemath: This needs to remain this way until the simple roomba estimator has better tracking of stopped versus moving states. I would like to merge this this way.

aftersomemath:
This is ready.

src/iarc7_motion/iarc_tasks/track_roomba_task.py
asaba96: I agree. Hopefully we will have something that can estimate the states soon.

PR #133, opened 2018-05-28
make roomba tasks use full frame ids

asaba96:


aftersomemath:
Doesn't this need an accompanying PR in abstract?

asaba96:
I rolled in the changes into the bump
Roomba PR

PR #135, opened 2018-06-08
made changes to land planner and crazyflie yaml file

MiscM:


param/low_level_motion_sim_1.5.yaml
```diff
@@ -70,10 +70,12 @@ takeoff_throttle_ramp_duration: 0.5
 # Rate at which to descent
 # must be negative
 descend_rate: -0.7
+cushion_rate: -.05
```
aftersomemath: Can these be set to the same values as the crazyflie was tested with?

aftersomemath: Also these parameters need to be added to the sim (not 1.5) and 2.0

src/LandPlanner.cpp
```diff
@@ -73,7 +81,8 @@ bool LandPlanner::prepareForTakeover(const ros::Time& time)
     }
 
     requested_height_ = transform.transform.translation.z;
-
+    // height determined by the ratio of landing accelerations
+    cushion_height_ = std::min(0.5 * (std::pow(descend_rate_,2)/cushion_acceleration_), requested_height_ * ( 1 - ( 1 / ( 1 - descend_acceleration_/cushion_acceleration_))));
```
aftersomemath: Can this be broken into multiple lines? Typically we try to limit line width to 80 characters.

src/LandPlanner.cpp
```diff
@@ -104,16 +113,31 @@ bool LandPlanner::getTargetMotionPoint(const ros::Time& time,
         return false;
     }
 
+    actual_height_ = transform.transform.translation.z;
```
aftersomemath: The height doesn't need to be stored off to a class member since no calculations depends on knowing the height from the last iteration. We can get rid of this and just use "transform.transform.translation.z" everywhere.

src/LandPlanner.cpp
aftersomemath: Is this equation for a worse case scenario of hitting the ground with zero velocity? Because with the cushion_rate limiter this will always be more conservative then it needs to be. However it won't be by a lot and I think its a great approximation.

param/low_level_motion_sim_1.5.yaml
aftersomemath: Can you add a quick comment to explain the usage of these parameters? It will be helpful to us when we forget what they mean the night before competition.

src/LandPlanner.cpp
```diff
@@ -24,20 +24,28 @@ LandPlanner::LandPlanner(
...
       startup_timeout_(ros_utils::ParamUtils::getParam<double>(
               private_nh,
               "startup_timeout")),
       update_timeout_(ros_utils::ParamUtils::getParam<double>(
               private_nh,
-              "update_timeout")),
+              "update_timeout")),                    
       uav_arm_client_(nh.serviceClient<iarc7_msgs::Arm>("uav_arm"))
 {
     landing_detected_subscriber_ = nh.subscribe("landing_detected",
```
aftersomemath: Can a ROS_ASSERT be added here to check that cushion_acceleration, cushion_rate_, descend_rate_, and descend_acceleration have the appropriate sign? I'd hate for a crash to occur because they were wrong. Cushion_rate and cushion_accel are particularly susceptible to being confused IMO.

include/iarc7_motion/LandPlanner.hpp
```diff
@@ -68,15 +68,19 @@ class LandPlanner
 
     LandState state_;
 
+
```
Liam-Berti: I don't like this

src/LandPlanner.cpp
```diff
@@ -149,8 +176,15 @@ bool LandPlanner::getTargetMotionPoint(const ros::Time& time,
...
     motion_point.motion_point.twist.linear.z = actual_descend_rate_;
 
-    if(actual_descend_rate_ > descend_rate_) {
-        motion_point.motion_point.accel.linear.z = descend_acceleration_;
+    if (transform.transform.translation.z > cushion_height_) {
+        if(actual_descend_rate_ > descend_rate_) {
+            motion_point.motion_point.accel.linear.z = descend_acceleration_;
+        }
+    }
+    else {
```
Liam-Berti: this should be }else{

include/iarc7_motion/LandPlanner.hpp
asaba96: I agree. Too much white space

src/LandPlanner.cpp
asaba96: I also agree

src/LandPlanner.cpp
```diff
@@ -72,8 +83,11 @@ bool LandPlanner::prepareForTakeover(const ros::Time& time)
         return false;
     }
 
-    requested_height_ = transform.transform.translation.z;
+    requested_height_ = transform.transform.translation.z; 
 
+    // height determined by the ratio of landing accelerations
+    cushion_height_ = std::min(0.5 * (std::pow(descend_rate_,2)/cushion_acceleration_),
```
aftersomemath: We need to look at the offset between the landing gear and the center of lift to calculate this properly as revealed by testing.

include/iarc7_motion/LandPlanner.hpp
amiller27: image

src/LandPlanner.cpp
```diff
@@ -24,26 +24,37 @@ LandPlanner::LandPlanner(
...
               private_nh,
-              "update_timeout")),
+              "update_timeout")),                    
       uav_arm_client_(nh.serviceClient<iarc7_msgs::Arm>("uav_arm"))
 {
     landing_detected_subscriber_ = nh.subscribe("landing_detected",
                                  100,
                                  &LandPlanner::processLandingDetectedMessage,
                                  this);
+    ROS_ASSERT_MSG((descend_rate_ <= 0) &&
```
amiller27: Can we make this four separate asserts?  Then we'll know immediately which one failed

src/LandPlanner.cpp
```diff
@@ -106,14 +120,27 @@ bool LandPlanner::getTargetMotionPoint(const ros::Time& time,
...
-                                          + (actual_descend_rate_
-                                          * (time - last_update_time_).toSec()));
+      // determines whether to speed up or slow down, depending on height
+        if (transform.transform.translation.z > cushion_height_) {
+            actual_descend_rate_ = std::max(descend_rate_,
+                                            actual_descend_rate_
+                                            + (descend_acceleration_
+                                            * (time - last_update_time_).toSec()));
+
+            requested_height_ = std::max(0.0, requested_height_
```
amiller27: Can we move `requested_height_` out of this `if..else` block?

src/LandPlanner.cpp
amiller27: What's the reasoning behind these two?  You don't need to go through all the math here, but there should be a high-level description somewhere of what assumptions you're starting with

PR #142, opened 2018-07-21
Don't cancel a goal that's not there

aftersomemath:


aftersomemath:
Fast creation and cancellation of a task. Needs more testing, its hard to reproduce.

